PR #5, opened 2022-10-13
Add preliminary device inventory

jgallagher:
This allows us to fetch a full list of

```rust
pub struct SpDevice {
    pub device: String,
    pub description: String,
    pub num_measurement_channels: u32,
    pub presence: DevicePresence,
}
```

from an SP, where `device`/`description`/`presence` are analogous to what `humility validate` shows today, and `num_measurement_channels` indicates which devices can report measurements (what we currently call "sensors" but are in the process of renaming).

This PR includes one commit (280c52c) to _remove_ the assumption that any trailing data sent in the remainder of a packet is prefixed by a 2-byte length, since we're now sending TLV-packed data in this case. After discussion with @andrewjstone I don't think the link prefix is necessary in the two cases we were using it (update chunks and serial console data), since the length is already implicit in the amount of data contained in the packet.

### DIFF
--- a/gateway-messages/src/sp_impl.rs
+++ b/gateway-messages/src/sp_impl.rs
@@ -304,9 +304,16 @@ fn encode_device_inventory<H: SpHandler>(
                 total_tlv_len += n;
                 out = &mut out[n..];
             }
+            Err(tlv::EncodeError::Custom(infallible)) => {
+                // A bit of type system magic here; our custom error type
+                // (`Infallible`) cannot be instantiated. We can
+                // provide an empty match to teach the type system that an
+                // `Infallible` (which can't exist) can this branch is
+                // unreachable without needing to explicitly panic.
+                match infallible {}
+            }
             // We checked the length above; this error isn't possible.
             Err(tlv::EncodeError::BufferTooSmall) => panic!(),
-            Err(tlv::EncodeError::Custom(infallible)) => match infallible {},
         }
 
         device_index += 1;
@@ -351,7 +358,7 @@ fn read_request_header(data: &[u8]) -> (u32, Result<&[u8], ResponseError>) {
 /// are handled by `read_request_header`) and calls `handler`.
 ///
 /// If the response kind needs to generate trailing data, returns `(Ok(_),
-/// Some(_)`, and our caller is resposnible for handling that generation.
+/// Some(_)`, and our caller is responsible for handling that generation.
 /// Otherwise, returns `(result, None)`.
 fn handle_message_impl<H: SpHandler>(
     sender: SocketAddrV6,
--- a/gateway-messages/src/sp_impl.rs
+++ b/gateway-messages/src/sp_impl.rs
@@ -30,7 +30,6 @@ use crate::UpdateChunk;
 use crate::UpdateId;
 use crate::UpdateStatus;
 use core::convert::Infallible;
-use core::mem;
 
 #[cfg(feature = "std")]
 use std::net::SocketAddrV6;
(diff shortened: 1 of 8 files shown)


gateway-messages/src/sp_impl.rs
```diff
@@ -38,6 +42,26 @@ pub struct SocketAddrV6 {
...
+/// Description of a device as reported as part of this SP's inventory.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub struct DeviceDescription<'a> {
+    pub device: &'a str,
+    pub description: &'a str,
+    pub num_measurement_channels: u32,
+    pub presence: DevicePresence,
+}
+
+impl From<DeviceDescription<'_>> for crate::DeviceDescription {
```
andrewjstone, line 53: I find this somewhat confusing, since they both have the same name. Later on, you alias `crate::DeviceDescription` as `DeviceDescriptionHeader`. I think it would be useful to just call it `crate::DeviceDescriptionHeader` to begin with. That's really all it's used for, as it doesn't contain enough information unserialized to do anything with.

gateway-messages/src/sp_impl.rs
```diff
@@ -209,10 +251,84 @@ pub fn handle_message<H: SpHandler>(
...
+                Ok(n)
+            },
+        ) {
+            Ok(n) => {
+                total_tlv_len += n;
+                out = &mut out[n..];
+            }
+            // We checked the length above; this error isn't possible.
+            Err(tlv::EncodeError::BufferTooSmall) => panic!(),
+            Err(tlv::EncodeError::Custom(infallible)) => match infallible {},
```
andrewjstone, line 310: I don't really understand what's going on here. Can you enlighten me just for learning purposes? Thanks!

gateway-messages/src/sp_impl.rs
```diff
@@ -249,42 +365,68 @@ fn read_request_header(data: &[u8]) -> (u32, Result<&[u8], ResponseError>) {
 
 /// Parses the remainder of a request (after `version` and `request_id`, which
 /// are handled by `read_request_header`) and calls `handler`.
+///
+/// If the response kind needs to generate trailing data, returns `(Ok(_),
+/// Some(_)`, and our caller is resposnible for handling that generation.
```
andrewjstone, line 355: typo: responsible

andrewjstone (approved):
Nice work @jgallagher.

gateway-messages/src/sp_impl.rs
jgallagher: Yeah, sorry, this is tricky enough I should put a comment here. `infallible` here has the type [`core::convert::Infallible`](https://doc.rust-lang.org/core/convert/enum.Infallible.html), "the error type for errors that can never happen". It's an uninhabitable enum, which means an instance of it can never exist. Here we appear to have one (in `infallible`), but we can attempt to `match` on it (with no arms, because it's an uninhabitable enum!). The compiler realizes that the match with no arms means `infallible` can't actually exist, which means from type system point of view it can be any type at all:

```rust
// This is fine.
fn infallible_to_u32(infallible: core::convert::Infallible) -> u32 {
    match infallible {}
}

// This is also fine.
fn infallible_to_string(infallible: core::convert::Infallible) -> String {
    match infallible {}
}

// Even this is fine.
fn infallible_to_anything<T>(infallible: core::convert::Infallible) -> T {
    match infallible {}
}
```

This is why it's really useful as an error type; you can unwrap without `.unwrap()`, and the compiler knows the error branch isn't possible:

```rust
fn unwrap_without_panic<T>(result: Result<T, core::convert::Infallible>) -> T {
    match result {
        Ok(t) => t,
        Err(infallible) => match infallible {},
    }
}
```

The actual code here is essentially doing this, but slightly more convoluted because it's only applicable for `EncodeError::Custom(_)`.

andrewjstone: Thanks for the explanation. I guess you could also use `()` as the error type and always unwrap, but that would generate more code and the branch could make it slower. So this is an optimization, right?

jgallagher: Right, this entire branch should not exist in the binary.

gateway-messages/src/sp_impl.rs
jgallagher: Good call; renamed in 8468175

PR #8, opened 2022-10-18
Minor updates to device description fields based on MGS integration

jgallagher:
1. Add an `SpComponent` that can be used for follow-up API calls related to each component
2. Replace `num_measurement_channels` with the more general `capabilities` bitmask.

### DIFF
--- a/gateway-messages/src/lib.rs
+++ b/gateway-messages/src/lib.rs
@@ -207,14 +207,27 @@ pub struct DeviceInventoryPage {
     Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, SerializedSize,
 )]
 pub struct DeviceDescriptionHeader {
+    pub component: SpComponent,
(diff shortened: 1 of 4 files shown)


gateway-messages/src/lib.rs
```diff
@@ -641,6 +658,43 @@ impl SpComponent {
...
     }
+
+    /// Interpret the component name as a human-readable string in a `const`
+    /// context, panicking if the string is not human readable.
+    ///
+    /// This function should only be used in const contexts when the caller
+    /// knows the component is valid (e.g., one of this type's associated
+    /// constants); for component names parsed or constructed at runtime, prefer
+    /// [`SpComponent::as_str()`] which performs runtime validation.
+    pub const fn const_as_str(&self) -> &str {
```
andrewjstone, line 669: Impressive!

andrewjstone (approved):

PR #9, opened 2022-10-24
Rework switch port config to allow specifying an interface

jgallagher:
Each management network is now configured by the triple `(listen_addr, discovery_addr, interface)`, where `listen_addr` and `discovery_addr` are expected to be identical for all ports (but may not be in tests where we're simulating a management network) and `interface` is expected to be specific to each port (but is optional, again to support simulating a management network). If an `interface` is provided, the `SingleSp` responsible for communication on that port will wait indefinitely for the interface to exist (checked via libc's `if_nametoindex()` function).

This PR also reworks how `SingleSp`s are initialized (and therefore how MGS startup works). Creating a new `SingleSp` can now only fail if there is an obvious misconfiguration in the toml file; it returns even before a UDP socket is bound, since it may need to wait an unknown amount of time for its `interface` to appear. Any method calls made on that `SingleSp` will fail while we're in the "starting up" phase. `SingleSp::wait_for_startup_completion()` provides a way for a caller to block until that startup is done, but MGS proper is _not_ expected to call that method - it only exists for use by `faux-mgs`.

### DIFF
--- a/gateway-sp-comms/src/lib.rs
+++ b/gateway-sp-comms/src/lib.rs
@@ -45,11 +45,11 @@ pub use timeout::Timeout;
 
 const DISCOVERY_MULTICAST_ADDR: Ipv6Addr =
     Ipv6Addr::new(0xff02, 0, 0, 0, 0, 0, 0, 1);
-const SP_MGS_PORT: u16 = 11111;
+const SP_PORT: u16 = 11111;
 
 /// Default address to discover an SP via UDP multicast.
 pub fn default_discovery_addr() -> SocketAddrV6 {
-    SocketAddrV6::new(DISCOVERY_MULTICAST_ADDR, SP_MGS_PORT, 0, 0)
+    SocketAddrV6::new(DISCOVERY_MULTICAST_ADDR, SP_PORT, 0, 0)
 }
 
 /// Default address to use when binding our local socket.
--- a/gateway-sp-comms/src/lib.rs
+++ b/gateway-sp-comms/src/lib.rs
@@ -18,6 +18,9 @@ mod management_switch;
 mod single_sp;
 mod timeout;
 
+use std::net::Ipv6Addr;
+use std::net::SocketAddrV6;
+
 pub use usdt::register_probes;
 
 pub mod error;
@@ -30,12 +33,30 @@ pub use management_switch::SpIdentifier;
 pub use management_switch::SpType;
 pub use management_switch::SwitchConfig;
 pub use management_switch::SwitchPortConfig;
+pub use management_switch::SwitchPortDescription;
 pub use single_sp::AttachedSerialConsole;
 pub use single_sp::AttachedSerialConsoleRecv;
 pub use single_sp::AttachedSerialConsoleSend;
 pub use single_sp::SingleSp;
 pub use single_sp::SpDevice;
 pub use single_sp::SpInventory;
-pub use single_sp::DISCOVERY_MULTICAST_ADDR;
 pub use timeout::Elapsed;
 pub use timeout::Timeout;
+
+const DISCOVERY_MULTICAST_ADDR: Ipv6Addr =
+    Ipv6Addr::new(0xff02, 0, 0, 0, 0, 0, 0, 1);
+const SP_MGS_PORT: u16 = 11111;
+
+/// Default address to discover an SP via UDP multicast.
+pub fn default_discovery_addr() -> SocketAddrV6 {
+    SocketAddrV6::new(DISCOVERY_MULTICAST_ADDR, SP_MGS_PORT, 0, 0)
+}
+
+/// Default address to use when binding our local socket.
+pub fn default_listen_addr() -> SocketAddrV6 {
+    // TODO: Currently the SP never tries to discover MGS, only MGS discovers
+    // SP, which means only SPs need to be listening on known ports. Can we bind
+    // the same port on all the vlan interfaces so in the future SPs could
+    // discover MGS if needed?
+    SocketAddrV6::new(Ipv6Addr::UNSPECIFIED, 0, 0, 0)
+}
--- a/gateway-sp-comms/src/single_sp.rs
+++ b/gateway-sp-comms/src/single_sp.rs
@@ -151,6 +151,7 @@ impl SingleSp {
         Self { state, inner_task, log }
     }
 
+    /// Block until all our local setup (see [`SingleSp::new()`] is complete.
     pub async fn wait_for_startup_completion(
         &self,
     ) -> Result<(), StartupError> {
@@ -159,6 +160,11 @@ impl SingleSp {
 
     /// Retrieve the [`watch::Receiver`] for notifications of discovery of an
     /// SP's address.
+    ///
+    /// This function only returns an error if startup has failed; if startup
+    /// has succeeded, always returns `Ok(_)` even if no SP has been discovered
+    /// yet (in which case the returned receiver will be holding the value
+    /// `None`).
     pub fn sp_addr_watch(
         &self,
     ) -> Result<&watch::Receiver<Option<(SocketAddrV6, SpPort)>>, StartupError>
--- a/gateway-sp-comms/src/single_sp.rs
+++ b/gateway-sp-comms/src/single_sp.rs
@@ -9,7 +9,9 @@
 use crate::communicator::ResponseKindExt;
 use crate::error::BadResponseType;
 use crate::error::SpCommunicationError;
+use crate::error::StartupError;
 use crate::error::UpdateError;
+use crate::SwitchPortConfig;
 use backoff::backoff::Backoff;
 use gateway_messages::tlv;
 use gateway_messages::version;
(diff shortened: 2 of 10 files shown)


gateway-sp-comms/src/lib.rs
```diff
@@ -30,12 +33,30 @@ pub use management_switch::SpIdentifier;
...
 pub use single_sp::SingleSp;
 pub use single_sp::SpDevice;
 pub use single_sp::SpInventory;
-pub use single_sp::DISCOVERY_MULTICAST_ADDR;
 pub use timeout::Elapsed;
 pub use timeout::Timeout;
+
+const DISCOVERY_MULTICAST_ADDR: Ipv6Addr =
+    Ipv6Addr::new(0xff02, 0, 0, 0, 0, 0, 0, 1);
+const SP_MGS_PORT: u16 = 11111;
```
andrewjstone, line 48: Nit: I find this naming slightly confusing. Based on the comment in `default_listen_addr()` it's the listen port for the SP. The MGS part throws me off though. Since we are talking in the context of MGS in this file anyway, maybe just call it `SP_PORT`.

gateway-sp-comms/src/single_sp.rs
```diff
@@ -113,44 +111,65 @@ impl Drop for SingleSp {
...
+    /// The local setup includes:
+    ///
+    /// 1. Waiting for an interface with the name specified by
+    ///    `config.interface` to exist (if `config.interface` is `Some(_)`). We
+    ///    determine this via `if_nametoindex()`. This step never fails, but
+    ///    will block forever waiting for the interface.
+    /// 2. Binding a UDP socket to `config.listen_addr` (with a `scope_id`
+    ///    determined by the previous step). If this bind fails (e.g., because
+    ///    `config.listen_addr` is invalid), the returned `SingleSp` will return
+    ///    a "UDP bind failed" error from all methods forever.
```
andrewjstone, line 128: As discussed in chat, I think this is fine, because bind should only fail if the address is misconfigured or already in use. Both of those are invariants that should not be violated and require manual intervention to sort out.

The only other time bind would fail is if the interface returned somehow wasn't valid. This likely also requires manual intervention. It may be worth checking with @Nieuwejaar if this interface will get removed/recreated during the lifetime of a booted scrimlet.

gateway-sp-comms/src/single_sp.rs
```diff
@@ -113,44 +111,65 @@ impl Drop for SingleSp {
...
+    pub async fn wait_for_startup_completion(
+        &self,
+    ) -> Result<(), StartupError> {
+        self.state.wait_for_startup_completion().await
     }
 
     /// Retrieve the [`watch::Receiver`] for notifications of discovery of an
     /// SP's address.
+    ///
+    /// This function only returns an error if startup has failed; if startup
```
andrewjstone, line 164: Nit: This code definitely works and makes sense, and the comments clarify its usage. I wouldn't have even noticed enough to make a suggestion without the comments here. However, that is kinda the point :) This is an implicit contract based on an Option being returned. Would it be better to make it a special enum type that reports whether an SP is discovered instead? Something like:

```rust
enum StartupState {
    DiscoverySucceeded(Inner),
    DiscoveryFailed
}
```

gateway-sp-comms/src/single_sp.rs
```diff
@@ -723,6 +743,12 @@ impl Inner {
     }
 
     async fn run(mut self) {
+        // If discovery fails (typically due to timeout, but also possible due
+        // to misconfiguration where we can't send packets at all), how long do
+        // we wait before retrying? If failure is due to misconfiguration, we
+        // will never succeed - how do we cope with that?
```
andrewjstone, line 749: I think this latter part is something we'll have to deal with via metrics/alerts at some point.

gateway-sp-comms/src/single_sp/startup.rs
```diff
@@ -0,0 +1,228 @@
...
+    // We're waiting for our configured network interface to exist.
+    WaitingForInterface(String),
+    // We're waiting for the result of attempting to bind to our listening
+    // interface.
+    WaitingToBind(SocketAddrV6),
+    // Startup is complete (although not necessarily successful!).
+    //
+    // TODO We currently treat failure to bind a socket as a terminal error,
+    // expecting it to be indicative of a configuration problem we have no
+    // recourse to solve. Is that correct, or do we need some way of recovering
```
andrewjstone, line 37: I think this is fine. @Nieuwejaar what do you think?

andrewjstone (approved):
Looks good. Just some minor nits from me.

gateway-sp-comms/src/single_sp.rs
Nieuwejaar: Under normal circumstances, I wouldn't expect it to.  However, if the sidecar gets powercycled or `dpd` or `tfportd` crash, then I would expect all of the `tfport` devices to be destroyed and recreated as part of the recovery.

andrewjstone: Oh boy. That could lead to interesting scenarios. Is it possible to have dpd or tfportd check to see if what it needs are already initialized, or must it recreate them? If the latter (or if that just makes things more rational form a viewpoint I'm not currently privy to), how would we go about detecting this. The interface ids could end up being wrong post bind, although maybe that's just a socket error that we retry on. I also wonder what happens if the ports try to get destroyed but because they are in use by the previously bound socket, illumos doesn't allow this. I'm not sure if this is possible or not, but it certainly seems like management here needs to be more robust to failures.

gateway-sp-comms/src/single_sp/startup.rs
Nieuwejaar: I believe this is true, if `WaitingForInterface` is satisfied by the appearance of an `ipadm` interface.  If that condition only means that there is a `dladm` link, then this isn't necessarily correct as you could be in a window between the creation of the link and the interface.  I suspect the condition is referring to the `ipadm`- interface, so this should be fine.

gateway-sp-comms/src/single_sp.rs
jgallagher: Can we tie the services together so that if `dpd` or `tfportd` are restarted, `mgs` is too? (If we can, would that be a reasonable way to address this?)

gateway-sp-comms/src/single_sp/startup.rs
jgallagher: I believe it is. The mechanics here are that we wait polling `if_nametoindex()` from libc until the interface exists. I just checked on a helios system: an interface existing in `dladm` only is not sufficient for `if_nametoindex()` to succeed; it needs to exist in `ipadm`.

gateway-sp-comms/src/single_sp.rs
andrewjstone: Ah great point! My Erlang [supervisor](https://www.erlang.org/doc/man/supervisor.html) brain should have thought of this. It looks like [SMF](https://illumos.org/man/7/smf) can do this via `restarters`.

Nieuwejaar: There probably are some specific cases where we could avoid rebuilding the world, but not in the general case.  We can't know how long we've been down and what may have changed in that time, so we can't know what should be cleaned up and what should be maintained.  We also can't know what information other daemons have about our state.

This has been a core design principle all along.  We have no mechanism for persisting/restoring any state at all between daemon launches.

andrewjstone: That makes sense. Thanks for the explanation Nils.

Nieuwejaar: In the event of the sidecar rebooting, I think our current plan is to restart the whole switch zone.

In the event of a daemon crash, If we think of `mgs` as being logically downstream of `dpd` and `tfportd`, then it makes sense to restart `mgs` when we restart the others.

Nieuwejaar: I should add that I don't have a firm opinion on whether it makes sense to think of `mgs` as downstream of `tfportd` or not.  Since `mgs` relies on interfaces that `tfportd` creates, then that feels like a reasonable mental model, but I don't know enough about `mgs` to say that with real certainty.

rmustacc: While restarting everything is an ok short term approach, it isn't what we want to do long term as that's pretty disruptive and unfortunate for clients and consumers. Most other HTTP services would basically try to return that something downstream isn't working as opposed to just not existing. While we have a lot of gotchas for getting to that state, I think being there is going to be a healthier place for trying to reliably build applications on top of these things and dealing with them coming and going. I do understand that it's not that simple for us to do and won't be there for some time possibly.

To put this in perspective, nexus isn't going to restart itself just because it loses connectivity with cockroach. I realize that's not really a fair analogy, but just something to think here. I realize and appreciate that none of this is as simple as a remote HTTP service coming and going.

gateway-sp-comms/src/single_sp.rs
jgallagher: Hmm, the option being `None` doesn't mean discovery has failed; it means discovery has not completed successfully (but might in the future). Discovery can't fail, except in cases of outright misconfiguration as discussed in other comments on this PR (bad socket address, failure to bind, etc), so this function can return:

* `Err(StartupError::*)` if we have not yet finished starting or failed to start (where "start" means waiting for the appropriate interface to show up and binding a UDP socket on that interface)
* `Ok(watch::Receiver<_>)` that currently has the value `None` - we've started up but have not yet discovered an SP; the `Receiver` will be sent a message when (if?) that changes
* `Ok(watch::Receiver<_>)` that currently has the value `Some(_)` - we've started up and discovered an SP; the `Receiver` will be sent a message if in the future we discover a different SP

andrewjstone: This makes sense. Ignore me!

PR #10, opened 2022-11-09
Rework message format to make it clearer that the SP can initiate requests

jgallagher:
Previously, the top-level message types included `Request` (only sent from MGS to SP) and `SpMessageKind::Response` (only sent from SP to MGS), strongly implying that requests always come from MGS and responses always come from SPs. That implication is incorrect; this PR reworks the top-level types to make it clear that either side can be the requester or the responder, via the `MessageKind` enum:

```rust
pub enum MessageKind {
    MgsRequest(MgsRequest),
    MgsResponse(MgsResponse),
    SpRequest(SpRequest),
    SpResponse(SpResponse),
}
```

The bulk of the PR deals with fallout from renaming or moving various message cases around; there is very little behavioral change other than shifting types around to fall within `MessageKind`, so please feel free to review briefly. Two behavioral changes include:

1. Bumping the protocol version number to 2; we still haven't shipped anything, but this is to quickly and obviously catch out-of-date SPs
2. Adding a check of the version number in `single_sp.rs` (this should've already been there!) - added in 99c3309


andrewjstone (approved):
Looks great!

PR #12, opened 2022-11-10
Add messages to get/set sled startup options

jgallagher:
`StartupOptions` is a bitflags struct that is copy/pasted from RFD 316 and the host/SP uart protocol. I _think_ it's reasonable for `gateway-messages` to include its own definitely (and require the SP to translate between the two structs) instead of trying to make `gateway-messages` depend on the host/SP messages crate or vice versa, but I could be talked out of it.

### DIFF
--- a/faux-mgs/src/main.rs
+++ b/faux-mgs/src/main.rs
@@ -86,8 +86,8 @@ enum Command {
     /// Ask SP for its current state.
     State,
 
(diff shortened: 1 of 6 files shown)


andrewjstone (approved):
LGTM. I don't have a real problem with the translation, but it would be nice if there was a shared definition that could be used between the two. I'm not going to bust your chops though :)

PR #21, opened 2022-11-28
Add SP version and RoT status to `SpState`

jgallagher:


jgallagher:
This is small enough I'm going to go ahead and merge, but if there's anything you think is missing from `RotState` we can come back and add it later.

PR #22, opened 2022-11-28
Flesh out ignition support

jgallagher:
We previously had some placeholder messages for ignition (based on an early RFD); now that ignition has landed in hubris, this fleshes out those messages to match it.

The bulk of this PR is adding new messages to get ignition state and link event state (for single targets or in bulk). Other notable changes:

* In `sp_impl.rs`, we now have a generic method to encode TLV triples at the end of a packet instead of having separate functions for each message type. Different messages have different serialization issues, so it takes an iterator that yields closures to perform the serialization; this feels a little awkward - suggestions very welcome.
* The TODO comments in `single_sp.rs` related to ignition only returning state for 35 targets (32 sleds, 2 psc, and the _other_ sidecar - not itself). This will require some thought/rework when we go to integrate into MGS proper, but I'm not sure how to handle that now, hence the TODOs.

### DIFF
--- a/gateway-messages/src/mgs_to_sp.rs
+++ b/gateway-messages/src/mgs_to_sp.rs
@@ -70,7 +70,7 @@ pub enum MgsRequest {
     BulkIgnitionLinkEvents {
         offset: u32,
     },
-    /// If `target` is `None`, clear events on all target (potentially
+    /// If `target` is `None`, clear events on all targets (potentially
     /// restricted by `transceiver_select`).
     ///
     /// If `transceiver_select` is none, clear events on all transceivers
--- a/gateway-messages/src/mgs_to_sp.rs
+++ b/gateway-messages/src/mgs_to_sp.rs
@@ -4,6 +4,7 @@
 
 //! Types for messages sent from MGS to SPs.
 
+use crate::ignition::TransceiverSelect;
 use crate::BadRequestReason;
 use crate::PowerState;
 use crate::SpComponent;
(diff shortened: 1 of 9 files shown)


gateway-messages/src/mgs_to_sp.rs
```diff
@@ -59,6 +62,23 @@ pub enum MgsRequest {
...
     },
+    /// Get ignition link events for a single target.
+    IgnitionLinkEvents {
+        target: u8,
+    },
+    /// Get ignition link events for all targets, starting at `offset`.
+    BulkIgnitionLinkEvents {
+        offset: u32,
+    },
+    /// If `target` is `None`, clear events on all target (potentially
```
andrewjstone, line 73: s/all target/all targets/

gateway-messages/src/sp_to_mgs/ignition.rs
```diff
@@ -0,0 +1,168 @@
...
+pub struct ReceiverStatus {
+    pub aligned: bool,
+    pub locked: bool,
+    pub polarity_inverted: bool,
+}
+
+#[derive(
+    Debug, Clone, Copy, PartialEq, Eq, SerializedSize, Serialize, Deserialize,
+)]
+pub struct Target {
```
andrewjstone, line 74: Nit: Everywhere else in this PR, `target` is a destination encoded as a `u8`. This is slightly confusing, although in context it would be clear. Maybe rename this type to `TargetState` to better reflect what it is.

andrewjstone (approved):
I actually really like the iterator solution. I think it's a practical abstraction here.
Nice jerb

PR #24, opened 2022-11-30
Update vsc7448/monorail counters

jgallagher:
This PR adds:

* A `ClearComponentStatus(_)` command (currently only implemented by the sidecar monorail, which it takes to mean "reset all counters")
* The new link flapping booleans to `PortCounters`

and renames "vsc7448" -> "monorail".


andrewjstone (approved):
Ship it!

PR #33, opened 2022-12-14
Make RotState reflect hubris state better

andrewjstone:
This matches the changes in [hubris#1009](https://github.com/oxidecomputer/hubris/pull/1009) but should merge first, once the hubris PR is ready for merge.

### DIFF
--- a/gateway-messages/src/sp_to_mgs.rs
+++ b/gateway-messages/src/sp_to_mgs.rs
@@ -153,13 +153,42 @@ pub struct SpState {
 #[derive(
     Debug, Clone, Copy, PartialEq, Eq, SerializedSize, Serialize, Deserialize,
 )]
-pub struct RotState {
+pub enum RotSlot {
+    A = 0,
+    B = 1,
+}
+
+#[derive(
+    Debug, Clone, Copy, PartialEq, Eq, Deserialize, Serialize, SerializedSize,
+)]
+pub struct RotImageDetails {
+    pub digest: [u8; 32],
     pub version: ImageVersion,
-    pub messages_received: u32,
-    pub invalid_messages_received: u32,
-    pub incomplete_transmissions: u32,
-    pub rx_fifo_overrun: u32,
-    pub tx_fifo_underrun: u32,
+}
+
+/// The boot time details dumped by Stage0 into Hubris on the RoT
+#[derive(
+    Debug, Clone, Copy, PartialEq, Eq, SerializedSize, Serialize, Deserialize,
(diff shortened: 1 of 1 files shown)


gateway-messages/src/sp_to_mgs.rs
```diff
@@ -149,13 +149,42 @@ pub struct SpState {
 #[derive(
     Debug, Clone, Copy, PartialEq, Eq, SerializedSize, Serialize, Deserialize,
 )]
-pub struct RotState {
+pub enum RotSlot {
+    A = 0,
```
jgallagher, line 153: What do we gain from the explicit `A = 0`, `B = 1`?

gateway-messages/src/sp_to_mgs.rs
```diff
@@ -149,13 +149,42 @@ pub struct SpState {
...
+    pub slot_a: Option<RotImageDetails>,
+    pub slot_b: Option<RotImageDetails>,
+}
+
+// TODO(AJS): Fill in with runtime state - i.e. updates that have completed before an RoT reset
+#[derive(
+    Debug, Clone, Copy, PartialEq, Eq, SerializedSize, Serialize, Deserialize,
+)]
+pub struct RotUpdateDetails {
+    pub boot_state: RotBootState,
```
jgallagher, line 180: Should `boot_state` be a member of `RotState` instead? This question is probably coming mostly from the fact that both `RotState` and `RotUpdateDetails` only have one member each; maybe you have ideas in mind for how each will be fleshed out and why this is placed where it is?

gateway-messages/src/sp_to_mgs.rs
andrewjstone: Nothing! I will remove it.

gateway-messages/src/sp_to_mgs.rs
andrewjstone: It's a good question. There's internally an RotUpdateDetails in hubris that is eventually going to contain the RotBootState and some runtime information as described in https://github.com/oxidecomputer/hubris/issues/1021. I was just leaving a placeholder for that. If you'd prefer I not do that I can change it for now. Either way is fine with me.

jgallagher: I don't have a strong preference either way if we're going to be changing it more soon anyway. I'm fine with leaving it as-is, just wanted to ask the question.

jgallagher (approved):

PR #34, opened 2022-12-15
Add InMemoryHostPhase2Provider, which parses version 2 phase2 image headers

jgallagher:
MGS will instantiate an instance of this provider with its capacity set to 1 or 2 (i.e., it will only keep one or two phase2 images around to serve to SPs), and its clients (nexus, wicketd) will upload images.

The larger change is that we no longer has the image, as that will not work with the upcoming image compression: we need to parse the image header to find the sha256 hash that the host (via the SP) will be requesting. In the future we will modify that header before sending it, but for now we forward it unmodified.


andrewjstone (approved):
No notes!

PR #45, opened 2023-01-18
Add currently-unused `ScopeIdCache`

jgallagher:
This is part of the larger rework for MGS to use a single shared UDP socket and then dispatch incoming packets based on their scope ID, which we then map back to an interface name. `ScopeIdCache` is unused other than its own tests in this PR, but is an easily-reviewable chunk on which additional work can be based.

We're going to do two lookups into small (~35) hash maps on _every_ incoming packet to MGS (one to map the packet's scope ID to an interface name, and a second to map the interface name to the handler for that interface). MGS isn't a particularly high-traffic service, but that still seemed like enough activity to warrant a couple minor optimizations: not using the default SipHash (since we have full control over the hash keys), and interning strings (to make cloning interface names cheap).

### DIFF
--- a/gateway-sp-comms/Cargo.toml
+++ b/gateway-sp-comms/Cargo.toml
@@ -1,13 +1,14 @@
 [package]
 name = "gateway-sp-comms"
 version = "0.1.0"
-edition = "2018"
+edition = "2021"
 license = "MPL-2.0"
 
 [dependencies]
 async-trait = "0.1"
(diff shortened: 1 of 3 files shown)


andrewjstone (approved):
Looks great. It took me a touch longer than I would have hoped to realize how the harness worked with Inner, but after I figured it out the tests were crystal clear and cover the functionality quite nicely!

PR #46, opened 2023-01-18
Add currently-unused SharedSocket

jgallagher:
This is the 2nd of three PRs working toward replacing "a `SingleSp` wraps a `UdpSocket`" with "MGS has one `UdpSocket` that is shared by all `SingleSp` handles". It introduces the `SharedSocket` type but does not use it anywhere (similar to the `ScopeIdCache` in #45, on which this PR builds). The final PR will integrate the two into the rest of `gateway-sp-comms`.

### DIFF
--- a/gateway-sp-comms/src/shared_socket.rs
+++ b/gateway-sp-comms/src/shared_socket.rs
@@ -1,0 +1,712 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+// Copyright 2023 Oxide Computer Company
+
+//! MGS communications share a single UDP socket that distinguishes which SP
+//! it's talking to based on the scope ID of the packet.
+//!
+//! SPs are logically identified by interface names, and scope IDs are mapped to
+//! those interface names.
+
+#![allow(dead_code)] // TODO remove once this is used
+
+use fxhash::FxHashMap;
+use gateway_messages::version;
+use gateway_messages::Header;
+use gateway_messages::Message;
+use gateway_messages::MessageKind;
+use gateway_messages::MgsError;
+use gateway_messages::MgsResponse;
+use gateway_messages::SpComponent;
+use gateway_messages::SpRequest;
+use gateway_messages::SpResponse;
+use slog::debug;
+use slog::error;
+use slog::o;
+use slog::warn;
+use slog::Logger;
+use std::collections::hash_map;
+use std::io;
+use std::net::Ipv6Addr;
+use std::net::SocketAddr;
+use std::net::SocketAddrV6;
+use std::sync::Arc;
+use std::time::Duration;
+use thiserror::Error;
+use tokio::net::UdpSocket;
+use tokio::sync::mpsc;
+use tokio::sync::mpsc::error::TrySendError;
+use tokio::sync::Mutex;
+use tokio::task::JoinHandle;
+
+use crate::error::HostPhase2Error;
+use crate::scope_id_cache::InterfaceError;
+use crate::scope_id_cache::Name;
+use crate::scope_id_cache::ScopeIdCache;
+use crate::HostPhase2Provider;
+use crate::SP_TO_MGS_MULTICAST_ADDR;
+
+#[derive(Debug, Error)]
+#[error("failed to bind to {addr}: {err})")]
+pub struct BindError {
+    pub addr: SocketAddrV6,
+    pub err: io::Error,
+}
+
+/// `SharedSocket` wraps a single UDP socket and allows multiple
+/// [`SingleSp`](crate::SingleSp) handles to use it, assuming each is assigned
+/// to a different underlying network interface.
+///
+/// This is designed to match the way management network VLAN interfaces are set
+/// up inside the switch zone: all interfaces are VLANs that sit on top of
+/// `tfportCPU0`, and they all share the _same_ IPv6 link-local address. This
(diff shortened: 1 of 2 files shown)


gateway-sp-comms/src/shared_socket.rs
```diff
@@ -0,0 +1,712 @@
...
+use crate::SP_TO_MGS_MULTICAST_ADDR;
+
+#[derive(Debug, Error)]
+#[error("failed to bind to {addr}: {err})")]
+pub struct BindError {
+    pub addr: SocketAddrV6,
+    pub err: io::Error,
+}
+
+/// `SharedSocket` wraps a single UDP socket and allows multiple
```
andrewjstone, line 58: Excellent comment. Thank you!

gateway-sp-comms/src/shared_socket.rs
```diff
@@ -0,0 +1,712 @@
...
+                }
+                Err(err) => err,
+            };
+
+            // UDP send failures are relatively rare; a likely case is that the
+            // interface we're using has been deleted (and hopefully
+            // recreated!). If we fail to send, we will try evicting our cached
+            // scope ID to see if it's changed; if it hasn't, return the send
+            // error.
+            if self.refresh_scope_id().await? {
```
andrewjstone, line 313: I don't know how this could possibly happen, but it would be weird to have the scope id continuously refresh and trigger an infinite loop. We could provide a max count, but that may be overkill.

gateway-sp-comms/src/shared_socket.rs
```diff
@@ -0,0 +1,712 @@
...
+
+    pub(crate) async fn recv(&mut self) -> SingleSpMessage {
+        // If `recv()` returns `None`, the `RecvHandler` task associated with
+        // the shared socket we're using has panicked; we'll propagate that
+        // panic.
+        self.recv.recv().await.expect("recv() task died")
+    }
+}
+
+// Trivial wrapper around `UdpSocket` that only exposes `send`: in our
```
andrewjstone, line 333: Love it!

gateway-sp-comms/src/shared_socket.rs
```diff
@@ -0,0 +1,712 @@
...
+        loop {
+            let (n, peer) = match self.socket.recv_from(&mut buf).await {
+                Ok((n, SocketAddr::V6(addr))) => (n, addr),
+                // We only use IPv6; we can't receive from an IPv4 peer.
+                Ok((_, SocketAddr::V4(_))) => unreachable!(),
+                Err(err) => {
+                    // Failing to recv _probably_ means our socket is
+                    // irrecoverably broken, but there isn't much we can do
+                    // about that from here. We'll sleep to avoid spamming the
+                    // logs, but someone will have no notice we're dead and
```
andrewjstone, line 436: s/no/to/

gateway-sp-comms/src/shared_socket.rs
```diff
@@ -0,0 +1,712 @@
...
+                Err(RecvError::InvalidMessageKind("MgsRequest"))
+            }
+            MessageKind::MgsResponse(_) => {
+                Err(RecvError::InvalidMessageKind("MgsResponse"))
+            }
+            MessageKind::SpRequest(SpRequest::HostPhase2Data {
+                hash,
+                offset,
+            }) => {
+                if !sp_trailing_data.is_empty() {
```
andrewjstone, line 550: Should we really ignore this, or is it better to drop the message as it's malformed. I feel like we shouldn't be liberal in what we accept here, but I'm open to argument.

andrewjstone (approved):
Looks good. Just minor comments.

gateway-sp-comms/src/shared_socket.rs
jgallagher: I don't think it hurts to put a max count, good idea. In practice I think the only way we could hit that is if something _else_ is in a (fast!) retry loop destroying and recreating interfaces, but if that's happening it seems better to log a "the scope ID has changed 5 times, we gave up" error than spin forever.

gateway-sp-comms/src/shared_socket.rs
jgallagher: My only argument is a fairly weak one: we don't check for (or even log, which is inconsistent with the choice here!) empty trailing data on other messages that we don't expect any; e.g., https://github.com/oxidecomputer/management-gateway-service/blob/9e9fc718edea6756954447acddc9c47166f9ec41/gateway-sp-comms/src/single_sp.rs#L195, https://github.com/oxidecomputer/management-gateway-service/blob/9e9fc718edea6756954447acddc9c47166f9ec41/gateway-sp-comms/src/single_sp.rs#L218, and many more (search for `, _data`). A few options:

* Remove this check and log and silently accept unexpected trailing data like we do in all those others
* Add checks to all the others to log unexpected trailing data
* Change this to fail if we get unexpected trailing data and add similar check+fail to all the others

Seems like 3 is probably the way to go, but is big enough it should be a followup. What do you think?

andrewjstone: I think 3 is indeed the correct path. I also don't feel this is critical and it definitely can be done in a followup.

gateway-sp-comms/src/shared_socket.rs
jgallagher: Fixed in f7b545598e20dca33228f09c77812caccc9f3d32

gateway-sp-comms/src/shared_socket.rs
jgallagher: Lifted this out to #49

Issue #51, opened 2023-01-20
want to send NMI to host system via faux-mgs


jclulow:
I would like to be able to send an NMI to the Gimlet in order to get into kmdb.

jgallagher:
also filed on the hubris side as https://github.com/oxidecomputer/hubris/issues/1068

jgallagher:
This is done as of #56

PR #56, opened 2023-01-26
Add APIs to send NMI and UART break

mkeeter:
This also adds UART break to the console, using the same key (Ctrl-A Ctrl-\) as `picocom`.

Needs to be tested on real hardware!

Closes https://github.com/oxidecomputer/hubris/issues/1068 and #51 once complete

TODO:
- [x] Plumb the GPIO NMI through `faux-mgs`

### DIFF
--- a/faux-mgs/src/usart.rs
+++ b/faux-mgs/src/usart.rs
@@ -99,6 +99,7 @@ pub(crate) async fn run(
                         send_tx.send(SendTxData::Break)
                             .await
                             .with_context(|| "failed to send data (task shutdown?)")?;
+                        println!("\n*** break sent ***");
                     }
                 }
 
--- a/faux-mgs/src/usart.rs
+++ b/faux-mgs/src/usart.rs
@@ -99,7 +99,7 @@ pub(crate) async fn run(
                         send_tx.send(SendTxData::Break)
                             .await
                             .with_context(|| "failed to send data (task shutdown?)")?;
-                        println!("\n*** break sent ***");
+                        println!("\n\r*** break sent ***\r");
                     }
                 }
 
--- a/faux-mgs/src/usart.rs
+++ b/faux-mgs/src/usart.rs
@@ -24,6 +24,7 @@ use crate::picocom_map::RemapRules;
 
 const CTRL_A: u8 = b'\x01';
 const CTRL_X: u8 = b'\x18';
+const CTRL_BACKSLASH: u8 = b'\x1c';
 
 pub(crate) async fn run(
     sp: SingleSp,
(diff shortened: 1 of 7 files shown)


faux-mgs/src/usart.rs
```diff
@@ -245,6 +263,13 @@ impl StdinOutBuf {
                         self.buf.push(c);
                     }
                 }
+                CTRL_BACKSLASH => {
+                    if self.in_prefix {
+                        return IngestResult::Break;
```
jgallagher, line 268: This early return will cause any characters remaining in `buf` to be silently discarded. (The same is true on the early exit return, but we're exiting anyway.) Maybe we should have a `mut result` that we modify in this loop, and always run it to completion before returning?

gateway-sp-comms/src/single_sp.rs
```diff
@@ -1070,6 +1070,20 @@ impl AttachedSerialConsoleSend {
...
+        let (tx, rx) = oneshot::channel();
+        self.inner_tx
+            .send(InnerCommand::Rpc(RpcRequest {
+                kind: MgsRequest::SerialConsoleBreak,
+                our_trailing_data: None,
+                response_tx: tx,
+            }))
+            .await
+            .unwrap();
+        rx.await.unwrap();
```
jgallagher, line 1084: Couple comments on this function:

1. There's a free `rpc` function that helps with some of the boilerplate (creating the oneshot, etc.).
2. We should check that the SP sent back the expected `SerialConsoleBreakAck`.

Combining both of those, I think this should be roughly:

```rust
        rpc(&self.inner_tx, MgsRequest::SerialConsoleBreak, None)
            .await
            .result
            .and_then(|(_peer, response, _data)| {
                response.expect_serial_console_break_ack()
            })
```

which will require adding `expect_serial_console_break_ack()` to `SpResponseExt`.

gateway-sp-comms/src/single_sp.rs
```diff
@@ -1070,6 +1070,20 @@ impl AttachedSerialConsoleSend {
...
+        Ok(())
+    }
 }
```
jgallagher, line 1087: We should also add a `SingleSp::send_host_nmi()` method and a corresponding `faux-mgs` subcommand to call it. It doesn't necessarily have to be a part of this PR but might make it more convenient to test.

jgallagher (commented):
Thanks for picking this up!

gateway-sp-comms/src/single_sp.rs
mkeeter: Yup, that was in my TODO list above; done in 89abdb393380383ef3c30e6a6e13757510eec73d

faux-mgs/src/usart.rs
mkeeter: That makes sense to me, done in 4aba91d7d4690abad4c1ada04610570f22cc1886

gateway-sp-comms/src/single_sp.rs
jgallagher: Ah sorry, missed the TODO list. My bad.

jgallagher (approved):

PR #58, opened 2023-02-01
Add message and plumbing to set IPCC key lookup values

jgallagher:


gateway-messages/src/mgs_to_sp.rs
```diff
@@ -97,6 +97,14 @@ pub enum MgsRequest {
...
     /// Send an NMI to the host by toggling a GPIO
     SendHostNmi,
+    /// Set the value for an IPCC `KeyLookup` request across the host/SP control
+    /// uart.
+    ///
+    /// The value is appended as trailing data. We currently assume that the max
+    /// length for any value fits into the trailing data of a single packet.
+    SetIpccKeyLookupValue {
+        key: u8,
+    },
```
jgallagher, line 107: Realized when starting to integrate this with MGS proper that I didn't add a "get" message here. Do we want one for parity? I'm not sure we actually need it for anything other than debugging; the expected flow is `MGS -(set)-> SP <-(get)- ioctl`.

PR #64, opened 2023-02-23
Add methods supporting host phase 2 progress reports

jgallagher:
This PR adds a handful of methods that expose what's going on when the SP requests host phase 2 images from MGS:

1. We remember the (1) most recent request from the SP; MGS and wicketd can use that as a proxy for how far along the MGS -> SP -> UART -> Host boot is (in terms of user reporting; this is inherently racy and a little inaccurate because we don't know that the SP/host has actually received and processed data this far into the file, and it assumes the host requests the file sequentially).
2. Add a method to discard the most recent request from the SP; this is primarily to allow for "I'm about to start recovery; if you still remember progress from some past recovery, forget about it".
3. Add a helper method to the host phase 2 trait to ask for the total size of an image by its ID.

### DIFF
--- a/gateway-sp-comms/src/host_phase2.rs
+++ b/gateway-sp-comms/src/host_phase2.rs
@@ -35,6 +35,12 @@ pub enum HostPhase2ImageError {
 
 #[async_trait]
 pub trait HostPhase2Provider: Send + Sync + 'static {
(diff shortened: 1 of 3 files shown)


andrewjstone (approved):
Ship it!

PR #66, opened 2023-03-09
Add API to read from caboose

mkeeter:


faux-mgs/src/main.rs
```diff
@@ -260,6 +260,13 @@ enum Command {
         value_path: PathBuf,
     },
 
+    /// Read a single key from the caboose
+    ReadCaboose {
+        /// 4-character ASCII string
+        #[arg(value_parser = parse_tlvc_key)]
```
andrewjstone, line 266: Love it.

gateway-sp-comms/src/sp_response_ext.rs
```diff
@@ -478,7 +481,18 @@ impl SpResponseExt for SpResponse {
             Self::SetIpccKeyLookupValueAck => Ok(()),
             Self::Error(err) => Err(CommunicationError::SpError(err)),
             other => Err(CommunicationError::BadResponseType {
-                expected: response_kind_names::SP_STATE,
+                expected: response_kind_names::SET_IPCC_KEY_LOOKUP_VALUE_ACK,
```
andrewjstone, line 484: Nice catch!

gateway-sp-comms/src/single_sp.rs
```diff
@@ -752,6 +752,16 @@ impl SingleSp {
             response.expect_set_ipcc_key_lookup_value_ack()
         })
     }
+
+    pub async fn get_caboose_value(&self, key: [u8; 4]) -> Result<Vec<u8>> {
```
andrewjstone, line 756: Presumably this is returning a `Vec<u8>` to allow for arbitrary data without having to change the message type.

andrewjstone (approved):
Looks great!

gateway-messages/src/sp_impl.rs
```diff
@@ -464,6 +469,10 @@ pub fn handle_message<H: SpHandler>(
             )
         }
         None => 0,
+        Some(OutgoingTrailingData::CabooseData(data)) => {
+            out[n..n + data.len()].copy_from_slice(data);
+            data.len()
+        }
```
jgallagher, line 475: Ridiculously tiny nitpick - would you mind shifting the `None` to be the last case? :sweat_smile:

gateway-messages/src/sp_impl.rs
```diff
@@ -786,6 +795,19 @@ fn handle_mgs_request<H: SpHandler>(
         MgsRequest::SetIpccKeyLookupValue { key } => handler
             .set_ipcc_key_lookup_value(sender, port, key, trailing_data)
             .map(|()| SpResponse::SetIpccKeyLookupValueAck),
+        MgsRequest::ReadCaboose { key } => {
+            handler.get_caboose_value(key).map(|data| {
+                if data.len() > crate::MIN_TRAILING_DATA_LEN {
```
jgallagher, line 800: This is overly conservative - the actual amount of space available for trailing data varies based on the size of the serialized header+message payload, and in the `CabooseValue`case that's quite small; `MIN_TRAILING_DATA_LEN` is the length left after the _largest_ possible message payload. I'm fine leaving this as-is for now; if we end up with caboose values with lengths in between this conservative restriction and what would actually fit (but still in a single packet) we could revisit it, but at that point we probably want to consider multiple packets anyway.

gateway-messages/src/sp_to_mgs.rs
```diff
@@ -101,6 +101,9 @@ pub enum SpResponse {
     SendHostNmiAck,
     SetIpccKeyLookupValueAck,
     ComponentSetAndPersistActiveSlotAck,
+
+    /// The size of the trailing caboose data
+    CabooseValue(u32),
```
jgallagher, line 106: I don't think we need the `u32` (and I think you're actually not using it on the MGS side even in this PR).

jgallagher (approved):
LGTM - just a couple small comments/suggestions.

gateway-messages/src/sp_to_mgs.rs
mkeeter: Good point, removed!

gateway-messages/src/sp_impl.rs
mkeeter: Sure, done

PR #68, opened 2023-03-13
Adopt transceivers-style protocol versioning backwards compatibility 

jgallagher:
This PR changes our requirement for an exact version match in the message header to allowing "any version >= our `MIN` version as long as we can still deserialize the message", along with unit tests to ensure future changes do not break the serialized form of any existing messages. This allows us to easily add new variants to all our hubpack-encoded enums (always at the end!), and bump the version without breaking compatibility with any MGS / SP instance that doesn't understand the new variant(s). See https://github.com/oxidecomputer/transceiver-control/pull/66 for more discussion / background on the choices here.

The vast majority of the diff on this PR is the (very tedious) set of unit tests for the encoding of every `{Mgs,Sp}{Request,Response}` variant. Feel free to skip the tests entirely, or at most just skim them for general structure.

### DIFF
--- a/gateway-messages/src/lib.rs
+++ b/gateway-messages/src/lib.rs
@@ -26,8 +26,31 @@ pub use sp_to_mgs::*;
 /// Maximum size in bytes for a serialized message.
 pub const MAX_SERIALIZED_SIZE: usize = 1024;
 
+/// Module specifying the minimum and current version of the MGS protocol.
+///
+/// Our primary mechanism for serializing requests and responses is enums
+/// encoded via hubpack. It is easy to extend these enums by adding new
+/// variants, but changing, reordering, or removing existing variants is
+/// (usually) a breaking change.
+///
+/// Our plan for versioning this protocol is simple: for as long as we can,
+/// leave `version::MIN` unchanged, and do not change, reorder, or remove
+/// existing variants. When we add new variants, increase `CURRENT. Both the SP
+/// and MGS will attempt to deserialize any message with a version that is at
+/// least `MIN`. If the deserialization fails and the message version is higher
+/// than `CURRENT`, we note a version mismatch error (with the expectation that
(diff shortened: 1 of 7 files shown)


gateway-messages/src/lib.rs
```diff
@@ -26,8 +26,35 @@ pub use sp_to_mgs::*;
...
+/// Module specifying the minimum and current version of the MGS protocol.
+///
+/// Our primary mechanism for serializing requests and responses is enums
+/// encoded via hubpack. It is easy to extend these enums by adding new
+/// variants, but changing, reordering, or removing existing variants is
+/// (usually) a breaking change.
+///
+/// Our plan for versioning this protocol is simple: for as long as we can,
+/// leave `version::MIN` unchanged, and do not change, reorder, or remove
+/// existing variants. When we add new variants, increase `CURRENT. Both the SP
```
mkeeter, line 38:
```suggestion
/// existing variants. When we add new variants, increase `CURRENT`. Both the SP
```

gateway-messages/src/sp_impl.rs
```diff
@@ -515,15 +518,19 @@ where
 /// Read the request header from the front of `data`, returning the message ID
 /// we should use in our response (pulled from the header if possible, or a
 /// sentinel if not) and either the remainder of the request data or an error.
-fn read_request_header(data: &[u8]) -> (u32, Result<&[u8], SpError>) {
+fn read_request_header(data: &[u8]) -> (Header, Result<&[u8], SpError>) {
```
mkeeter, line 521: This could be a _little_ more strongly typed:
```suggestion
fn read_request_header(data: &[u8]) -> (Option<Header>, Result<&[u8], SpError>) {
```
(then, the caller would get to decide what fake message ID to use)

mkeeter (approved):

PR #70, opened 2023-03-17
New messages 'ResetComponent{Prepare,Trigger}' to allow reset of RoT

lzrd:
This enables RoT reset during update.

The merge will need to be coordinated with the merge of hubris merge of reset-with-intent.
The image preference after reset {None, ImageA, ImageB, TransientA, TransientB} will need to be plumbed after this.

### DIFF
--- a/gateway-messages/src/lib.rs
+++ b/gateway-messages/src/lib.rs
@@ -141,6 +141,35 @@ impl From<UpdateId> for uuid::Uuid {
     }
 }
 
+/// Affect boot image selection after reset.
+#[derive(
+    Debug,
+    Clone,
+    Copy,
+    PartialEq,
+    Eq,
+    Hash,
+    SerializedSize,
+    Serialize,
+    Deserialize,
+)]
+pub enum ResetIntent {
+    /// Perform a normal reset without altering boot image selection.
+    Normal,
+    /// The selected slot is made persistent.
+    Persistent,
+    /// The selected slot is selected on the next reset without regard
+    /// to the persistent selection.
+    Transient,
+    // TODO:
+    // Reset and transition from Prod-signed to Dev-signed images.
+    // There must be a valid firmware installation before the reset is performed.
+    // When changes to boot policy are one-way,
+    // to be performed and a signed per-device per-boot session challenge
+    // must accompany the request. Details TBD.
+    ExpensiveAndIrrevocableProdToDev = 86,
+}
+
 /// Identifier for a single component managed by an SP.
 #[derive(
     Clone, Copy, PartialEq, Eq, Hash, SerializedSize, Serialize, Deserialize,
--- a/gateway-messages/src/mgs_to_sp.rs
+++ b/gateway-messages/src/mgs_to_sp.rs
@@ -7,6 +7,7 @@
 use crate::ignition::TransceiverSelect;
 use crate::BadRequestReason;
 use crate::PowerState;
+use crate::ResetIntent;
 use crate::SpComponent;
 use crate::UpdateId;
 use hubpack::SerializedSize;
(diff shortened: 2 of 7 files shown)


gateway-messages/src/mgs_to_sp.rs
```diff
@@ -48,8 +48,8 @@ pub enum MgsRequest {
     },
     GetPowerState,
     SetPowerState(PowerState),
-    ResetPrepare,
-    ResetTrigger,
+    ResetPrepare(SpComponent),
+    ResetTrigger(SpComponent),
```
jgallagher, line 52: CI is failing because this makes a backwards-incompatible change to an existing message, which we shouldn't do - in this case, pushing this update would prevent us from resetting existing SPs that don't have this change.

I think the easiest thing to do here is to just add `ResetComponent{Prepare,Trigger}(SpComponent)` variants to the end of `MgsRequest`, and probably add a corresponding response ack message to the end of `SpResponse`.

A question while we're here - for the SP reset is split into prepare/trigger so MGS can get an ack to the "prepare" message. Is that necessary for resetting components?

jgallagher (changes requested):

gateway-messages/src/mgs_to_sp.rs
lzrd: I assumed that the prepare was to be able to quiesce things or reject, e.g. wrong power state. Is the intent of prepare to just ping the SP?

I'm ok with not making a breaking change. A new message and then later deprecation of the old if appropriate is reasonable.

jgallagher: > I assumed that the prepare was to be able to quiesce things or reject, e.g. wrong power state. Is the intent of prepare to just ping the SP?

At the moment yes - `control-plane-agent` doesn't really have any means to quiesce things. It could check the power state but doesn't do that either. If we might want to do either of those for component resets (which seems likely), :+1: on keeping the prepare/trigger pair for the new variants.

lzrd: NM. I see the details of prepare.

gateway-messages/src/lib.rs
```diff
@@ -141,6 +141,35 @@ impl From<UpdateId> for uuid::Uuid {
...
+    PartialEq,
+    Eq,
+    Hash,
+    SerializedSize,
+    Serialize,
+    Deserialize,
+)]
+pub enum ResetIntent {
+    /// Perform a normal reset without altering boot image selection.
+    Normal = 1,
```
jgallagher, line 158: Do we need explicit values on this and `ExpensiveAndIrrevocableProdToDev`? They aren't used by hubpack when serializing so seem a little misleading.

gateway-sp-comms/src/single_sp.rs
```diff
@@ -762,6 +763,73 @@ impl SingleSp {
...
+        };
+        let response = rpc(
+            &self.cmds_tx,
+            MgsRequest::ResetComponentTrigger { component, slot, intent },
+            data,
+        )
+        .await;
+        match response.result {
+            Ok((_addr, _response, _data)) => Ok(()),
+            Err(other) => Err(other),
```
jgallagher, line 830: Two questions here:

* For the `Ok` case, can we check the response type (e.g., `response.expect_sys_reset_component_trigger_ack()`)?
* For the `Err` case, could we check if `component` is `SP_ITSELF` and if so, convert the error we expect to get if the SP has reset into an `Ok(())` response for our caller the way the existing `reset_trigger` function does?

jgallagher (commented):
This looks good to me, just a couple minor questions.

gateway-messages/src/lib.rs
lzrd: That "=1" can go away.

jgallagher (approved):
LGTM - thanks!

PR #77, opened 2023-04-03
Plumb switch default image

andrewjstone:
This is the MGS side of the code required for RoT boot selection as well resetting the RoT. A corresponding hubris PR will be out later today.

This is all @lzrd's code and it looks great to me. I have no proposed changes. I tested it out using corresponding hubris code and verified by reading the right CFPA registers via `humility readmem`. The boot versions change as expected.

I'm opening the PR just to move this forward and get @jgallagher's review since MGS is largely his baby.

### DIFF
--- a/gateway-messages/src/mgs_to_sp.rs
+++ b/gateway-messages/src/mgs_to_sp.rs
@@ -129,6 +129,16 @@ pub enum MgsRequest {
 
     SerialConsoleKeepAlive,
 
+    /// Reset a specific component
+    /// SP_ITSELF and ROT are supported
+    ResetComponentPrepare {
+        component: SpComponent,
+    },
+    ResetComponentTrigger {
+        component: SpComponent,
+    },
+
+    /// Change boot image selection on reset or power-on.
     SwitchDefaultImage {
         component: SpComponent,
         slot: SlotId,
--- a/gateway-messages/src/mgs_to_sp.rs
+++ b/gateway-messages/src/mgs_to_sp.rs
@@ -7,7 +7,9 @@
 use crate::ignition::TransceiverSelect;
 use crate::BadRequestReason;
 use crate::PowerState;
+use crate::SlotId;
 use crate::SpComponent;
+use crate::SwitchDuration;
 use crate::UpdateId;
 use hubpack::SerializedSize;
 use serde::Deserialize;
@@ -126,6 +128,12 @@ pub enum MgsRequest {
     },
 
     SerialConsoleKeepAlive,
+
+    SwitchDefaultImage {
+        component: SpComponent,
+        slot: SlotId,
+        duration: SwitchDuration,
+    },
 }
 
 #[derive(
--- a/gateway-messages/src/sp_impl.rs
+++ b/gateway-messages/src/sp_impl.rs
@@ -227,12 +227,6 @@ pub trait SpHandler {
         port: SpPort,
     ) -> Result<(), SpError>;
 
-    fn reset_prepare(
-        &mut self,
-        sender: SocketAddrV6,
-        port: SpPort,
-    ) -> Result<(), SpError>;
-
     /// Number of devices returned in the inventory of this SP.
     fn num_devices(&mut self, sender: SocketAddrV6, port: SpPort) -> u32;
 
@@ -1092,14 +1086,6 @@ mod tests {
         }
 
         fn serial_console_break(
-            &mut self,
-            _sender: SocketAddrV6,
-            _port: SpPort,
-        ) -> Result<(), SpError> {
-            unimplemented!()
-        }
-
-        fn reset_prepare(
             &mut self,
             _sender: SocketAddrV6,
             _port: SpPort,
--- a/gateway-messages/src/sp_impl.rs
+++ b/gateway-messages/src/sp_impl.rs
@@ -25,6 +25,7 @@ use crate::MgsRequest;
 use crate::MgsResponse;
 use crate::PowerState;
 use crate::SerializedSize;
+use crate::SlotId;
 use crate::SpComponent;
 use crate::SpError;
 use crate::SpPort;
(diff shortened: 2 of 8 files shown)


gateway-messages/src/mgs_to_sp.rs
```diff
@@ -126,6 +128,22 @@ pub enum MgsRequest {
...
+    ResetComponentTrigger {
+        component: SpComponent,
+    },
+
+    /// Change boot image selection on reset or power-on.
+    SwitchDefaultImage {
+        component: SpComponent,
+        slot: SlotId,
+        duration: SwitchDuration,
+    },
```
jgallagher, line 146: Can you expand the `gateway_messages::tests::versioning::v2::mgs_request()` test to cover these new variants? (This is the test that ensures future changes don't break the on-the-wire form of any pre-existing variants.)

gateway-messages/src/sp_to_mgs.rs
```diff
@@ -106,6 +106,9 @@ pub enum SpResponse {
     CabooseValue,
 
     SerialConsoleKeepAliveAck,
+    ResetComponentPrepareAck,
+    ResetComponentTriggerAck,
+    SwitchDefaultImageAck,
```
jgallagher, line 111: Please add coverage for these to `gateway_messages::tests::versioning::v2::sp_response()`

gateway-messages/src/sp_to_mgs.rs
```diff
@@ -471,6 +474,15 @@ pub enum SpError {
...
     ImageBoardMismatch,
+    /// Received a `ResetComponentTrigger` request without first receiving a
+    /// `ResetComponentPrepare` request. This can be used to detect a successful
+    /// reset on the SP_ITSELF. Not used for the ROT where the SP observes
+    /// the RoT reset and can report success.
+    ResetComponentTriggerWithoutPrepare,
+    /// There will be policy violations for some requests:
+    ///   - No image in SlotId (what suitability checks are needed?)
+    ///   - Lower epoch than current in SlotId
+    SwitchDefaultImageError(u32),
```
jgallagher, line 485: These should also be covered by `gateway_messages::tests::versioning::v2::sp_response()` (there's a loop inside it that covers all the `SpError` variants)

gateway-sp-comms/src/single_sp.rs
```diff
@@ -762,6 +764,86 @@ impl SingleSp {
...
+    /// needed the reset for has happened (e.g., checking the SP's version, in
+    /// the case of updates).
+    pub async fn reset_component_prepare(
+        &self,
+        component: SpComponent,
+    ) -> Result<()> {
+        debug!(
+            self.log,
+            "sending ResetComponentPrepare component:{component:?}"
+        );
```
jgallagher, line 786: I don't think we need the two `debug!` logs in this method - all messages sent and received are already logged (at the `trace!` level) by `Inner`'s handlers for rpc requests.

gateway-sp-comms/src/single_sp.rs
```diff
@@ -762,6 +764,86 @@ impl SingleSp {
...
+    pub async fn switch_default_image(
+        &self,
+        component: SpComponent,
+        slot: SlotId,
+        duration: SwitchDuration,
+    ) -> Result<()> {
+        let response = self
+            .rpc(MgsRequest::SwitchDefaultImage { component, slot, duration })
+            .await;
+        match response {
```
jgallagher, line 840: Style nit - I think this could be written as an `.and_then` instead of an explicit match (e.g., see `ignition_command()`).

gateway-messages/src/sp_impl.rs
```diff
@@ -358,6 +360,31 @@ pub trait SpHandler {
...
+
+    fn reset_component_prepare(
+        &mut self,
+        sender: SocketAddrV6,
+        port: SpPort,
+        component: SpComponent,
+    ) -> Result<(), SpError>;
+
+    // On success, this method will return unless the reset
+    // affects the SP_ITSELF.
```
mkeeter, line 372: This is confusing to me, because we already have `reset_prepare` and `reset_trigger`.  Are there cases where we'd use `reset_component_trigger(... SpComponent::SP_ITSELF)` instead of `reset_trigger`?

Adding a comment to that effect would be helpful.

gateway-messages/src/sp_impl.rs
```diff
@@ -656,7 +683,8 @@ fn handle_mgs_request<H: SpHandler>(
     let trailing_data = match &kind {
         MgsRequest::UpdateChunk(_)
         | MgsRequest::SerialConsoleWrite { .. }
-        | MgsRequest::SetIpccKeyLookupValue { .. } => leftover,
+        | MgsRequest::SetIpccKeyLookupValue { .. }
+        | MgsRequest::ResetComponentTrigger { .. } => leftover,
```
mkeeter, line 687: What trailing data do we expect in `ResetComponentTrigger`

andrewjstone: Good catch! I removed it.

gateway-sp-comms/src/single_sp.rs
andrewjstone: removed.

gateway-sp-comms/src/single_sp.rs
andrewjstone: fixed.

gateway-messages/src/sp_impl.rs
andrewjstone: After discussing with @jgallagher I removed `reset_prepare` and `reset_trigger` altogether.

andrewjstone: I realize I missed a few spots. I'm cleaning it up.

andrewjstone:
I just successfully updated the SP and RoT with the corresponding changes made to reset_prepare/trigger

jgallagher (approved):

PR #78, opened 2023-04-05
Format update UUID

bnaecker:
Tiny fix I noticed when resolving the issue that lead to https://github.com/oxidecomputer/hubris/pull/1263. We called `.to_string()` rather than `format!()` leading to log messages containing things like `update {update_id} aborted` instead of a formatted string.


jgallagher (approved):
Nice catch, thank you!

PR #82, opened 2023-04-15
Add Sprot related errors

andrewjstone:
This adds error types that can be returned by sprot. Each of the errors is actually from a distinct hubris API, but all are combined into the top-level `SprotError`.  This is supported in hubris via the new `Complex` errors in idol. A bunch of `From` implementations exist in https://github.com/oxidecomputer/hubris/pull/1279 to convert the underlying hubris errors to the `gateway-messages` errors. 

This PR puts sprot related errors into both `SpError` and `RotError`. It's unclear if we actually want to maintain `RotError` at all. Currently it's only used in when getting state from the RoT. 

The errors in hubris map 1:1 to the gateway-messages errors right now, with the addition of a currently unused `Unknown(u32)` variant in each gateway-messages error. This is there to support upgrades. The plan for upgrade
is that new variants will be added to the hubris error and the `From` implementation in hubris will put the index of the new error into the `Unknown` variant in the gateway-messages version. This gives us a way to figure out what error occurred by looking at the from implementation. This allows updated SPs and RoTs to work with older MGSs. This situation, however is expected to be only temporary. The new variant will be added to the gateway-messages type and then MGSs will be upgraded. Once gateway-messages has the new variant, hubris will be updated to pull in the newer version of gateway-messages and the `From` implementation will be updated, and the Unknown mapping will be removed as it is no longer needed. 

We also plan to ensure that MGS can talk to older versions of the SP, and so if we can update MGS first we will. Then multiple updates will not be required to get MGS and the SP in sync. In this case we'd just add the new variants to MGS and hubris simultaneously and never use the `Unknown` gateway-messages variant.

In all cases, there needs to be an update mechanism capable of making these decisions about what is possible to update and what is not. 

This PR also:
 * Bumps hubpack to the latest version. Thanks to @cbiffle for pushing a new update with the latest human readable changes.
 * Makes serde use `default-features` = false. This is because when gateway-messages was pulled into the `sprot-api`, a ton of compilation errors related to `std` not being available resulted. This fixed that, and seems to have not caused other issues.

### DIFF
--- a/gateway-messages/src/sp_to_mgs.rs
+++ b/gateway-messages/src/sp_to_mgs.rs
@@ -893,6 +893,19 @@ impl std::error::Error for IpccKeyLookupValueError {}
 )]
 pub enum RotError {
     MessageError { code: u32 },
+
+    // --------------------------------------
+    // *** That new hotness below here ***
+    // --------------------------------------
+    //
+    // New nested variants, one for each hubris API are below We will likely
+    // create a new Error variant so we can deprecate some of redundant
+    // variants above.
+    Sprot(SprotProtocolError),
+    Spi(SpiError),
+    Sprockets(SprocketsError),
+    Update(UpdateError),
+    Dump(DumperError),
 }
 
 impl fmt::Display for RotError {
@@ -901,6 +914,11 @@ impl fmt::Display for RotError {
             Self::MessageError { code } => {
                 write!(f, "SP/RoT messaging error: {code}")
             }
+            Self::Sprot(e) => write!(f, "sprot: {}", e),
+            Self::Spi(e) => write!(f, "spi: {}", e),
+            Self::Sprockets(e) => write!(f, "sprockets: {}", e),
+            Self::Update(e) => write!(f, "update: {}", e),
+            Self::Dump(e) => write!(f, "dump: {}", e),
         }
     }
 }
--- a/gateway-messages/src/sp_to_mgs.rs
+++ b/gateway-messages/src/sp_to_mgs.rs
@@ -499,6 +499,7 @@ pub enum SpError {
     Update(UpdateError),
     Dump(DumperError),
 }
+
 impl fmt::Display for SpError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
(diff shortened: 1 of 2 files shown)


gateway-messages/src/sp_to_mgs.rs
```diff
@@ -590,6 +603,262 @@ impl fmt::Display for SpError {
...
+    StartReadFailed,
+    ReadFailed,
+    BadDumpAreaHeader,
+    WriteFailed,
+    HeaderReadFailed,
+    FailedToHalt,
+    FailedToResume,
+    FailedToResumeAfterFailure,
+    RegisterReadFailed,
+    ServerRestarted,
```
jgallagher, line 655: Do we need to expose `ServerRestarted`? In general I think I'd expect these to be handled locally within the SP via retries, unless a task restarting means it panicked and lost some state that means it can't proceed. If we want to keep this one, we should probably rename it `TaskRestarted` to avoid confusion about what "server" means.

(Same comment on the other `ServerRestarted` variants later in this file.)

jgallagher (commented):
Apologies for how tedious this request is, but to ensure we don't break these messages moving forward, could you:

1. Bump [version::CURRENT](https://github.com/oxidecomputer/management-gateway-service/blob/caea9e57378c8d1989206c04d12cfe81f5b8aada/gateway-messages/src/lib.rs#L66) from 3 to 4?
2. Add a [v4](https://github.com/oxidecomputer/management-gateway-service/blob/caea9e57378c8d1989206c04d12cfe81f5b8aada/gateway-messages/tests/versioning/mod.rs#L9-L10) test module that checks the serialized forms of all the new errors?

gateway-messages/src/sp_to_mgs.rs
andrewjstone: I think we have to expose it. It's an error that can be returned from the API on non-idempotent operations. While it shouldn't happen N retries in a row, it might. I support changing the name though. Will do that.

andrewjstone: > Apologies for how tedious this request is, but to ensure we don't break these messages moving forward, could you:
> 
> 1. Bump [version::CURRENT](https://github.com/oxidecomputer/management-gateway-service/blob/caea9e57378c8d1989206c04d12cfe81f5b8aada/gateway-messages/src/lib.rs#L66) from 3 to 4?
> 2. Add a [v4](https://github.com/oxidecomputer/management-gateway-service/blob/caea9e57378c8d1989206c04d12cfe81f5b8aada/gateway-messages/tests/versioning/mod.rs#L9-L10) test module that checks the serialized forms of all the new errors?

Done in [060f6fa](https://github.com/oxidecomputer/management-gateway-service/pull/82/commits/060f6fa2371244cb1cd4d744e0ab3aa10d43f3a4)

PR #83, opened 2023-04-17
Bump hubpack and fix `serde`

mkeeter:
`hubpack v0.1.2` is released, so let's use it!

`serde` needs `default-features = false`; otherwise, feature unification will enable default features (including `std`) in the Hubris build!


jgallagher (approved):

andrewjstone:
Ha, I had these same fixes in https://github.com/oxidecomputer/management-gateway-service/pull/82

PR #86, opened 2023-05-01
Remove "switch default image" in favor of "set component active slot"

jgallagher:
This does not remove the underlying message as we don't want to worry about breaking compatibility, but changes the SP handler to no longer need a `switch_default_image` (we call `set_component_active_slot` if we receive the `SwitchDefaultImage` message from an out-of-date faux-mgs).


andrewjstone (approved):
Thanks for deduping this.

PR #87, opened 2023-05-03
Add data sent to host phase2 progress

jgallagher:
This required some refactoring to send the progress reports _after_
we've requested data from the host phase2 provider and sent data to the
SP, which probably makes them more accurate anyway! Previously we would
update our progress even if the request from the SP was nonsense or we
failed to reply to it; now we only update progress after we know we've
successfully sent data to an SP (although of course we don't know for
sure that it received it).

The primary motivator for this is correctness in wicket: prior to this PR, progress
for delivery of the trampoline image looks like it stalls at 99.9%, because we
were reporting the _start_ of the last data request from the SP, but we really
want to report the _end_. That requires the changes in this PR: we need to know
how much data we sent in response to the SP's request.

### DIFF
--- a/gateway-sp-comms/Cargo.toml
+++ b/gateway-sp-comms/Cargo.toml
@@ -26,7 +26,7 @@ usdt.workspace = true
 uuid.workspace = true
 zip.workspace = true
 
-gateway-messages.workspace = true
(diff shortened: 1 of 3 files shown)


andrewjstone (approved):
Looks great!

PR #88, opened 2023-05-03
Use hubtools to read hubris archive

jgallagher:
This also allows to add a preflight check that the caboose BORD value matches our target SP.

Also checks in `Cargo.lock`, which we probably should have been doing for a long time since `faux-mgs` lives here.


mkeeter (approved):
This looks good!

Consider writing a TODO in the RoT update function to add a similar caboose check there.

jgallagher:
> Consider writing a TODO in the RoT update function to add a similar caboose check there.

There is no explicit RoT update function (yet!). :sweat_smile: There will be in the next PR, and it will definitely have this TODO.

PR #91, opened 2023-05-12
Add support for reading an individual component's caboose

mkeeter:
There's a bunch of plumbing here and a corresponding Hubris PR that I will open shortly.

Briefly, this adds the `ReadComponentCaboose` message, which requests a component + slot + key (effectively deprecating the older `ReadCaboose` message).

The `CabooseValue` – used to serialize trailing data after the packet message – is split into three flavors:
- A local slice (`&'static [u8]`)
- A value in the caboose on the other SP flash bank
- A value in the RoT's caboose

There's a bunch of new code to handle this in `sp_impl.rs`; it's particularly gnarly because reading the caboose could fail (e.g. if the RoT refuses to answer us).

As a drive-by fix, I also made `SpComponent` deserialize nicely, to remove a bunch of boilerplate conversions.

### DIFF
--- a/faux-mgs/src/main.rs
+++ b/faux-mgs/src/main.rs
@@ -174,7 +174,8 @@ enum Command {
 
     /// Get or set the active slot of a component (e.g., `host-boot-flash`).
     ComponentActiveSlot {
-        component: String,
+        #[clap(value_parser = parse_sp_component)]
+        component: SpComponent,
         #[clap(short, long, value_name = "SLOT", help = "set the active slot")]
         set: Option<u16>,
         #[clap(
@@ -193,10 +194,16 @@ enum Command {
     Inventory,
 
     /// Ask SP for details of a component.
-    ComponentDetails { component: String },
+    ComponentDetails {
+        #[clap(value_parser = parse_sp_component)]
+        component: SpComponent,
+    },
 
     /// Ask SP to clear the state (e.g., reset counters) on a component.
-    ComponentClearStatus { component: String },
+    ComponentClearStatus {
+        #[clap(value_parser = parse_sp_component)]
+        component: SpComponent,
+    },
 
     /// Attach to the SP's USART.
     UsartAttach {
@@ -246,19 +253,24 @@ enum Command {
         #[clap(long)]
         allow_multiple_update: bool,
 
-        component: String,
+        #[clap(value_parser = parse_sp_component)]
+        component: SpComponent,
         slot: u16,
         image: PathBuf,
     },
 
     /// Get the status of an update to the specified component.
-    UpdateStatus { component: String },
+    UpdateStatus {
+        #[clap(value_parser = parse_sp_component)]
+        component: SpComponent,
+    },
 
     /// Abort an in-progress update.
     UpdateAbort {
         /// Component with an update-in-progress to be aborted. Omit to abort
         /// updates to the SP itself.
-        component: String,
+        #[clap(value_parser = parse_sp_component)]
+        component: SpComponent,
         /// ID of the update to abort.
         update_id: Uuid,
     },
@@ -289,6 +301,23 @@ enum Command {
         key: [u8; 4],
     },
 
+    /// Read a single key from the caboose
+    ReadComponentCaboose {
+        /// Component from which to read; must be `sp` or `rot`
+        #[clap(short, long, value_parser = parse_sp_component)]
+        component: SpComponent,
+
+        /// Target slot from which to read the caboose.
+        ///
+        /// The SP accepts `active` or `inactive`; the RoT accepts `A` or `B`
+        #[clap(short, long)]
+        slot: Option<String>,
+
+        /// 4-character ASCII string
+        #[arg(value_parser = parse_tlvc_key)]
+        key: [u8; 4],
+    },
+
     /// Instruct the SP to reset.
     Reset,
 
@@ -296,7 +325,10 @@ enum Command {
     ///
     /// This command is implemented for the component "rot" but may be
     /// expanded to other components in the future.
-    ResetComponent { component: String },
+    ResetComponent {
+        #[clap(value_parser = parse_sp_component)]
+        component: SpComponent,
+    },
 
     /// Controls the system LED
     SystemLed {
@@ -339,6 +371,11 @@ fn parse_tlvc_key(key: &str) -> Result<[u8; 4]> {
     }
 
     Ok(key.as_bytes().try_into().unwrap())
+}
+
+fn parse_sp_component(component: &str) -> Result<SpComponent> {
+    SpComponent::try_from(component)
+        .map_err(|_| anyhow!("invalid component name: {}", component))
 }
 
 #[derive(Debug, Clone, Copy)]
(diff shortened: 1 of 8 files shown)


faux-mgs/src/main.rs
```diff
@@ -289,14 +301,34 @@ enum Command {
...
 
+    /// Read a single key from the caboose
+    ReadComponentCaboose {
+        /// Component from which to read; must be `sp` or `rot`
+        #[clap(short, long, value_parser = parse_sp_component)]
+        component: SpComponent,
+
+        /// Target slot from which to read the caboose.
+        ///
+        /// The SP accepts `active` or `inactive`; the RoT accepts `A` or `B`
```
andrewjstone, line 312: I really like the specific targets here, but last time these were added, in `switch_default_image` (removed in favor of using existing calls like `set_component_active_slot`), both @jclulow and @jgallagher complained of the inconsistency with all the other commands that take slot directly as a `u16`. Personally, I'd prefer changing those to be more specific as well, but I"m not sure others feel the same.

faux-mgs/src/main.rs
```diff
@@ -341,6 +373,11 @@ fn parse_tlvc_key(key: &str) -> Result<[u8; 4]> {
     Ok(key.as_bytes().try_into().unwrap())
 }
 
+fn parse_sp_component(component: &str) -> Result<SpComponent> {
+    SpComponent::try_from(component)
+        .map_err(|_| anyhow!("invalid component name: {component}"))
+}
```
jgallagher, line 379: Nice :+1:

faux-mgs/src/main.rs
jgallagher: This is probably a "how much time do you spend with the RoT" question. Josh and I don't spend much with it, and are used to using 0/1 for host slots (even though those are also named A/B in some contexts) and 0 (vacuously) for SP updates. I'll make a suggestion to try to appease everyone in the match below.

faux-mgs/src/main.rs
```diff
@@ -1097,6 +1107,44 @@ async fn run_command(
                 Ok(Output::Lines(vec!["done".to_string()]))
             }
         }
+        Command::ReadComponentCaboose { component, slot, key } => {
+            let slot = match (component, slot.as_deref()) {
+                (SpComponent::SP_ITSELF, Some("active") | None) => 0,
```
jgallagher, line 1112: Can we make all of these arms accept the underlying integer too? So

```suggestion
                (SpComponent::SP_ITSELF, Some("active") | Some("0") | None) => 0,
```

etc?

gateway-messages/src/lib.rs
```diff
@@ -237,6 +237,16 @@ impl<'de> Deserialize<'de> for SpComponent {
     }
 }
 
+impl std::fmt::Display for SpComponent {
```
jgallagher, line 240: Does this need a `#[cfg(feature = "std")]` guard to not break hubris?

gateway-messages/src/sp_impl.rs
```diff
@@ -350,10 +351,18 @@ pub trait SpHandler {
...
+        component: SpComponent,
+        slot: u16,
         key: [u8; 4],
-    ) -> Result<&'static [u8], SpError>;
+    ) -> Result<CabooseValue, SpError>;
+
+    fn copy_caboose_value_into(
+        &self,
+        value: CabooseValue,
+        out: &mut [u8],
```
jgallagher, line 364: I assume the implementer can assume `out.len()` matches what it is described by `value.len()`? Would it ever be possible that `value.len()` describes a max length, but we don't know the exact length until we get into this method (and therefore we should return a `usize` of the number of bytes written to `out`)?

gateway-messages/src/sp_impl.rs
```diff
@@ -493,11 +506,31 @@ pub fn handle_message<H: SpHandler>(
...
-            out[n..n + data.len()].copy_from_slice(data);
-            data.len()
+            // Because we didn't know the encoded length of the packet until
+            // now, we have to call back into the handler and copy caboose data
+            // into the trailing region.
+            //
+            // Unfortunately, unlike every other possible form of trailing data,
+            // this is faillible!  We check for errors here and re-serialize the
+            // whole chunk of packet data if there's a failure.
+            let len = data.len();
```
jgallagher, line 516: Maybe worth adding to the comment that we know `out[n..][..len]` won't panic because we already checked that `len <= crate::MIN_TRAILING_DATA_LEN`.

gateway-messages/src/sp_impl.rs
```diff
@@ -493,11 +506,31 @@ pub fn handle_message<H: SpHandler>(
...
+                let response = Message {
+                    header: Header { version: version::CURRENT, message_id },
+                    kind: MessageKind::SpResponse(SpResponse::Error(e)),
+                };
+
+                // We know `response` is well-formed and fits into `out`, so we
+                // can unwrap serialization.
+                n = hubpack::serialize(&mut out[..], &response).unwrap();
+            } else {
+                n += len;
```
jgallagher, line 529: Tiny nit - I have a mild preference for the "success" case coming first. How do you think this would look as a `match` instead of `if let ... else ...`? (My preference for a `match` might be stronger if `copy_caboose_value_into` returned a length that we needed to add to `n`.)

gateway-sp-comms/src/single_sp.rs
```diff
@@ -811,6 +811,25 @@ impl SingleSp {
                 response.expect_component_action_ack()
             })
     }
+
+    pub async fn read_component_caboose(
```
jgallagher, line 815: We probably don't want to remove `SingleSp::get_caboose_value()` (the old getter) immediately, but it's maybe worth a comment that we should remove it at some point once all the SPs we care about have been updated. Similarly, there's one use of a raw `MgsRequest::ReadCaboose` in `single_sp/update.rs`; we can't change that one yet either (because it prevent updating SPs that don't know about per-component caboose messages), but it could also use a comment that it should use the new request variant at some point?

faux-mgs/src/main.rs
jgallagher: Hah, you couldn't see my comment because it was in draft until seconds ago, but that's my exact suggestion, yeah.

faux-mgs/src/main.rs
mkeeter: Sure, done in 3f568f32327a5906bfbeafac497291c63a441d36

gateway-messages/src/lib.rs
mkeeter: Oddly, this didn't _seem_ to break Hubris, but I switched to `core::fmt::Display` for safety (76d0a9879fc6711e5230dc9e0f6f9e95552c5bd3)

gateway-messages/src/sp_impl.rs
mkeeter: Hmmm, this suggests that `copy_caboose_value_into` should take a bare `addr: u32` instead of a `value: CabooseValue`, to avoid this ambiguity.

faux-mgs/src/main.rs
mkeeter: Yeah, John agrees with that plan, so I did it below.

gateway-messages/src/sp_impl.rs
mkeeter: Ah, that won't work, because `CabooseValue` is an `enum`, not a simple position + length.  I'll add some commentary here to explain the invariants.

gateway-messages/src/sp_impl.rs
mkeeter: Sure, done in a09ea3743d8727e87512ebd86e248d7ff3c5484c

andrewjstone (approved):
Thanks for all the cleanup @mkeeter. This looks great.

gateway-sp-comms/src/single_sp.rs
mkeeter: Good point, added some comments in 6834e9d3b91b591d6d061e090276731354699730

jgallagher (approved):
> Thanks for all the cleanup @mkeeter. This looks great.

Seconded!

PR #92, opened 2023-05-17
[gateway-messages] Introduce SpStateV2

andrewjstone:
We introduce a new response for SpState requests. By modifying the expected response in gateway-sp-comms, we can receive both the original `SpState` and `SpStateV2` responses enclosed in a new `VersionedSpState` response. Faux-mgs just prints this, but MGS can return the combined results and allow upstream callers to decide what to do. This is a general strategy for message evolution, and in some cases MGS can only return the latest version by filling in optional fields with `None` if a new field is the only thing added to the new version of the struct.

Additionally, the strategy used here makes things easy on control-plane- agent as it only has to return the latest version of the type that it knows about.

### DIFF
--- a/faux-mgs/src/main.rs
+++ b/faux-mgs/src/main.rs
@@ -27,6 +27,7 @@ use gateway_sp_comms::SharedSocket;
 use gateway_sp_comms::SingleSp;
 use gateway_sp_comms::SpComponentDetails;
 use gateway_sp_comms::SwitchPortConfig;
+use gateway_sp_comms::VersionedSpState;
 use gateway_sp_comms::MGS_PORT;
 use serde_json::json;
 use slog::info;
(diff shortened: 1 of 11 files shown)


gateway-messages/src/sp_to_mgs.rs
```diff
@@ -189,7 +208,13 @@ pub struct RotBootState {
     pub slot_b: Option<RotImageDetails>,
 }
 
-// TODO(AJS): Fill in with runtime state - i.e. updates that have completed before an RoT reset
+/// The currently active boot slot, as well as persistent and transient boot
```
mkeeter, line 211: This docstring doesn't seem to describe an empty `struct RotBootInfo {}`

gateway-messages/tests/versioning/v6.rs
```diff
@@ -0,0 +1,99 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! The tests in this module check that the serialized form of messages from MGS
+//! protocol version 5 have not changed.
```
mkeeter, line 6:
```suggestion
//! protocol version 6 have not changed.
```

gateway-messages/tests/versioning/v6.rs
```diff
@@ -0,0 +1,99 @@
...
+//!
+//! If a test in this module fails, _do not change the test_! This means you
+//! have changed, deleted, or reordered an existing message type or enum
+//! variant, and you should revert that change. This will remain true until we
+//! bump the `version::MIN` to a value higher than 4, at which point these tests
+//! can be removed as we will stop supporting v5.
```
mkeeter, line 12:
```suggestion
//! bump the `version::MIN` to a value higher than 6, at which point these tests
//! can be removed as we will stop supporting v6.
```

(also, if you wouldn't mind fixing the block comment in `v5.rs`, it's also wrong)

faux-mgs/src/main.rs
```diff
@@ -723,38 +724,74 @@ async fn run_command(
                 return Ok(Output::Json(serde_json::to_value(state).unwrap()));
```
jgallagher, line 724: I don't know if @jclulow is consuming JSON for calling `state` yet, but it changes in this PR (even though this line didn't change). I'll let Josh say how painful that will be for him?

gateway-messages/src/sp_to_mgs.rs
```diff
@@ -163,6 +166,22 @@ pub struct SpState {
     pub rot: Result<RotState, RotError>,
 }
```
jgallagher, line 167: Maybe add a comment on `SpState` that it is quasi-deprecated and will only be returned by SPs with images older than the introduction of `SpStateV2`?

gateway-sp-comms/src/sp_response_ext.rs
```diff
@@ -255,12 +256,13 @@ impl SpResponseExt for SpResponse {
         }
     }
 
-    fn expect_sp_state(self) -> Result<SpState> {
+    fn expect_sp_state(self) -> Result<VersionedSpState> {
         match self {
-            Self::SpState(state) => Ok(state),
+            Self::SpState(state) => Ok(VersionedSpState::V1(state)),
```
jgallagher, line 261: Do you want to rename `SpState` -> `SpStateV1`? Hubpack doesn't care about the names so that wouldn't affect serialization, but might be slightly nicer to look at? Similarly for the names of the old structs/enums that now have `...V2` successors.

gateway-sp-comms/src/sp_response_ext.rs
```diff
@@ -255,12 +256,13 @@ impl SpResponseExt for SpResponse {
         }
     }
 
-    fn expect_sp_state(self) -> Result<SpState> {
+    fn expect_sp_state(self) -> Result<VersionedSpState> {
```
jgallagher, line 259: This did come together nicely :+1:

gateway-sp-comms/src/sp_response_ext.rs
mkeeter: (I second this motion)

andrewjstone: Done in https://github.com/oxidecomputer/management-gateway-service/pull/92/commits/1879755cbe398073368338d11717a8ac8dc7fd93

gateway-messages/src/sp_to_mgs.rs
andrewjstone: Leftover. Removed in https://github.com/oxidecomputer/management-gateway-service/pull/92/commits/1879755cbe398073368338d11717a8ac8dc7fd93

gateway-messages/tests/versioning/v6.rs
andrewjstone: Fixed in https://github.com/oxidecomputer/management-gateway-service/pull/92/commits/1879755cbe398073368338d11717a8ac8dc7fd93

gateway-messages/tests/versioning/v6.rs
andrewjstone: Fixed in https://github.com/oxidecomputer/management-gateway-service/pull/92/commits/1879755cbe398073368338d11717a8ac8dc7fd93

gateway-messages/src/sp_to_mgs.rs
andrewjstone: Fixed in [8d4cea8](https://github.com/oxidecomputer/management-gateway-service/pull/92/commits/8d4cea8b6861908414ee76f068cf212622eda10d)

faux-mgs/src/main.rs
andrewjstone: @jclulow says we can do what we need to do :)

andrewjstone:
@mkeeter @jgallagher I think this thing is ready to go!

jgallagher (approved):
Ship it!

PR #94, opened 2023-05-22
Add `UpdateStatus::RotError`

jgallagher:
This will allow us to remove the placeholder ["update error 9999" in control-plane-agent](https://github.com/oxidecomputer/hubris/blob/a4b14431b2ae77ded2f55c0116cf98dad4002d38/task/control-plane-agent/src/update/rot.rs#L125-L140).

While I was here, I unified the identical enums `SlotId` and `RotSlot` into `RotSlotId`. This changes the type of some past versions, but does not change their serialized form, since both enums were identical.


andrewjstone (approved):
Thanks for fixing this!

PR #96, opened 2023-06-01
Time out requests even during trampoline delivery

### LINKED ISSUE #95: MGS requests to some SPs became unresponsive during rack2 mupdate
Yesterday while MUpdating ~10 sleds in rack2 concurrently, all 10 were in the "trampoline image delivery" phase, and wicketd appeared to no longer be getting progress reports (i.e., the manifestation in the UI was that the trampoline delivery was "stuck"). The trampoline was still being delivered, which we confirmed by attaching to the console of the gimlets in question. Their SPs were also still responsive, as we were able to get on the console and send other commands via `faux-mgs` on jeeves. The fact that the trampoline delivery was still making progress indicates that the network connection between MGS and the SPs was fine.

Looking at the wicketd and MGS logs for one of the SPs in question, we see a period of almost 20 minutes during which _every_ wicketd request to MGS associated with that SP timed out. There were no timeouts before or after that period. At the end of the period, we see a few of these logs in MGS:

```
01:13:34.325Z WARN 411bd717-38da-40e6-8d5b-7c3306e7fb2f (ManagementSwitch): RPC requester disappeared while waiting for response
01:13:34.325Z WARN 411bd717-38da-40e6-8d5b-7c3306e7fb2f (ManagementSwitch): RPC requester disappeared while waiting for response
01:13:34.372Z WARN 411bd717-38da-40e6-8d5b-7c3306e7fb2f (ManagementSwitch): RPC requester disappeared while waiting for response
01:13:34.420Z WARN 411bd717-38da-40e6-8d5b-7c3306e7fb2f (ManagementSwitch): RPC requester disappeared while waiting for response
01:13:34.426Z WARN 411bd717-38da-40e6-8d5b-7c3306e7fb2f (ManagementSwitch): RPC requester disappeared while waiting for response
```

This time coincides with the end of the trampoline image delivery - in wicket, we saw it jump from the percentage at which it got stuck to 100%.

Requests to an SP ultimately flow throw a tokio channel into [this arm of a tokio::select!](https://github.com/oxidecomputer/management-gateway-service/blob/da51768e10dbe260cc6302bcfd2ac379e275a313/gateway-sp-comms/src/single_sp.rs#L1532-L1540); requests timing out (but all other external signs indicating everything was healthy) strongly suggest the task running this loop was stuck somewhere else.

We collected a core file but I'm not aware of a good way to investigate stuck tokio tasks from a core. Reading through the code, I suspect the problem is in rpc_call_one_attempt, specifically [this timeout](https://github.com/oxidecomputer/management-gateway-service/blob/da51768e10dbe260cc6302bcfd2ac379e275a313/gateway-sp-comms/src/single_sp.rs#L1848-L1856). In that loop, we are attempting to:

1. Send a single request to the SP
2. If we get a `Busy` response, we will back off and retry the request.
3. If we get an out-of-band message (in particular, a note that the SP has requested host phase 2 data, exactly what was happening during the mupdate), we ignore it and repeat the loop _without_ resending the request.

My hypothesis is:

1. We sent a single request to the SP. This request was _lost_ somehow (possibly the SP's incoming packet queue dropped it because there were too many requests in flight; this is entirely possible and largely the reason we have retry loops at all).
2. The timeout linked above never fired, because we continued to get notifications about host phase 2 data.
3. Once the trampoline image fully delivered, the timeout finally fired, and we broke out of the loop.

If this is correct, I think it should be relatively easy to fix: we are effectively restarting the timeout any time we get a `HostPhase2Request` or `SerialConsole` packet, and we should not do that.

jgallagher:
Prior to this change, we could fail to time out waiting for a response from an SP if we were still receiving out-of-band messages from it. During trampoline delivery, this could manifest as the SP becoming unresponsive until trampoline delivery completes; see https://github.com/oxidecomputer/management-gateway-service/issues/95.

Fixes #95.

### DIFF
--- a/gateway-sp-comms/src/single_sp.rs
+++ b/gateway-sp-comms/src/single_sp.rs
@@ -63,7 +63,6 @@ use tokio::sync::oneshot;
 use tokio::sync::watch;
 use tokio::task::JoinHandle;
 use tokio::time;
-use tokio::time::timeout;
 use tokio::time::Instant;
 use uuid::Uuid;
 
@@ -1836,23 +1835,29 @@ impl<T: InnerSocket> Inner<T> {
         // (e.g., a serial console relay).
         let mut resend_request = true;
 
+        // We want a resettable timeout, so we'll use an `Interval`. We only
+        // care about the first tick (see the `select!` below); if it fires,
+        // we've timed out and will give up.
+        //
+        // Whenever we send the request, we reset this interval. Critically, we
+        // can loop _without_ resending (and therefore without resetting this
+        // interval) - this allows us to still time out even if we're getting a
+        // steady stream of out-of-band messages.
+        let mut timeout = tokio::time::interval(self.per_attempt_timeout);
+
         loop {
             if resend_request {
                 self.socket_handle.send(serialized_request).await?;
+                timeout.reset();
             }
 
             // Reset our default policy of resending requests if we iterate on
             // this loop.
             resend_request = true;
 
-            let message = match timeout(
-                self.per_attempt_timeout,
-                self.socket_handle.recv(),
-            )
-            .await
-            {
-                Ok(result) => result,
-                Err(_elapsed) => return Ok(None),
+            let message = tokio::select! {
+                result = self.socket_handle.recv() => result,
+                _ = timeout.tick() => return Ok(None),
             };
 
             let (peer, header, response, sp_trailing_data) = match message {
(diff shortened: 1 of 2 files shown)


gateway-sp-comms/src/single_sp.rs
```diff
@@ -2052,3 +2057,110 @@ mod probes {
...
+                     (we expected a timeout after 200ms",
+                    start.elapsed(),
+                );
+            }
+            Ok(Ok(Some(value))) => panic!("unexpected response {value:?}"),
+            Ok(Err(err)) => panic!("unexpected error {err}"),
+            Err(_elapsed) => {
+                panic!(
+                    "rpc_call_one_attempt failed to time out \
+                     (expected timeout after 200ms, waited 2000ms)"
```
jgallagher, line 2161: Without the corresponding change to timeout handling, this is the test failure I saw, which matches the behavior we saw on rack2 (where our "2 second timeout" here is playing the role of wicketd - seeing timeouts when sending requests to specific SPs).

andrewjstone (approved):
Looks like a solid fix, and a nice test!

Did you run the test to ensure that it fails without your change?

jgallagher:
I did indeed!

PR #98, opened 2023-06-22
Add error type to mirror the AttestError from Hubris.

flihp:


flihp:
I'm in the process of getting the attest task merged (https://github.com/oxidecomputer/hubris/issues/1422#issuecomment-1603342431) and will be following that up with the sprot work to expose certs. That'll require getting this new error type merged here.

jgallagher (commented):
This looks fine, but I'm wondering why this is being added here / now - is there an MGS request that could produce this error?

A nitpick: with additions to the enums, we also need to (a) bump the current version number and (b) add tests for the serialization format. See #91 for an example (bumps the version to 5, adds a `tests/versioning/v5.rs` that covers the new variant). But I'd hold off on doing this until the question above is answer. :)

flihp:
Hmmm from my (brief) reading of the sprot code this appeared to be a necessary part of the error propagation, maybe you can help clear this up:

The `SprotError` type is used to send all possible errors between the RoT & SP sprot servers. To enable xfer of the error type from the `attest-api` crate I add a new variant to the `SprotError` type here: https://github.com/oxidecomputer/hubris/blob/49763c7bf592d25cfdcfcdad3b19554b919f6735/drv/sprot-api/src/error.rs#L32. There's a `From<SprotError> for SpError` impl in this same file which, if I'm understanding this correctly, means we must be able to to transform this new variant to the target type `SpError`. This is a type from the `gateway-messages` crate: https://github.com/oxidecomputer/management-gateway-service/blob/146a687f7413bfe580869bb6017f3bfe8b4710b1/gateway-messages/src/sp_to_mgs.rs#L452 and adding the types from this PR was my way of satisfying this requirement.

So I agree that adding these types would make a lot more sense if there was a way for them to be produced in response to a request from MGS. I just couldn't come up with another way of satisfying the `From<SprotError> for SpError` impl w/o these new types. I'm very much open to other approaches if you know of alternatives.

andrewjstone:
> Hmmm from my (brief) reading of the sprot code this appeared to be a necessary part of the error propagation, maybe you can help clear this up:
> 
> The `SprotError` type is used to send all possible errors between the RoT & SP sprot servers. To enable xfer of the error type from the `attest-api` crate I add a new variant to the `SprotError` type here: https://github.com/oxidecomputer/hubris/blob/49763c7bf592d25cfdcfcdad3b19554b919f6735/drv/sprot-api/src/error.rs#L32. There's a `From<SprotError> for SpError` impl in this same file which, if I'm understanding this correctly, means we must be able to to transform this new variant to the target type `SpError`. This is a type from the `gateway-messages` crate:
> 
> https://github.com/oxidecomputer/management-gateway-service/blob/146a687f7413bfe580869bb6017f3bfe8b4710b1/gateway-messages/src/sp_to_mgs.rs#L452
> 
> and adding the types from this PR was my way of satisfying this requirement.
> So I agree that adding these types would make a lot more sense if there was a way for them to be produced in response to a request from MGS. I just couldn't come up with another way of satisfying the `From<SprotError> for SpError` impl w/o these new types. I'm very much open to other approaches if you know of alternatives.

Sorry for coming late to the party. I've been in my own trust quorum related world. If attest errors are never exposed through MGS and can't be generated by MGS, then we shouldn't expose such errors to MGS. Regardless of that though, the attest error also shouldn't be part of `SprotError`. Bundling stuff other than SPI inside sprot error was a mistake I made that we've rectified with another pattern. See https://github.com/oxidecomputer/hubris/pull/1435/files#r1240040957 for details.

flihp:
Incredibly useful info @andrewjstone. Thanks for the pointer & clarification.

andrewjstone:
> Incredibly useful info @andrewjstone. Thanks for the pointer & clarification.

You're welcome. Sorry for the confusion!

Issue #49, opened 2023-01-20
Reject packets from SP with unexpected trailing data


jgallagher:
See https://github.com/oxidecomputer/management-gateway-service/pull/46#discussion_r1081847669 for context; if we receive a packet from an SP that deserializes successfully as a `Message` but contains unexpected trailing data, we currently do one of

a) ignore it
b) log it

but we should real return an error and discard the malformed packet.

PR #105, opened 2023-08-18
Clean up `expect_*` functions

mkeeter:
This PR uses the ✨ Power of Macros ✨  to reduce `SpResponse` unpacking boilerplate.

There are two big changes:
- Using `strum_macros::IntoStaticStr` to automatically implement variant-to-`str` conversions (in lower snake case).  This eliminates the table of strings in `response_kind_names`.
- Replacing `trait SpResponseExt` with free functions, which are mostly autogenerated by a macro.  This avoids duplication (declaring the functions in both the trait and implementation), dramatically reduces human-written code, and eliminates the possibility of human errors.

There's one small logical change.  The existing code unwrapped the result of `expect_caboose_value()`, i.e.
```rust
result.result.map(|(_peer, response, data)| {
    response.expect_caboose_value().unwrap();
    data
})
```
I've switched to propagating this error instead, because unwraps are bad.

--------------------------------------------------------------------------------

## Two open questions for reviewers

Easy:  Should we rename `sp_response_ext.rs` now that it's not actually an extension trait? If so, what should we call it?  (my answer is "yes, `sp_response_expect.rs`")

Hard: Should the `expect_*` functions take a tuple of `(SocketAddrV6, SpResponse, Vec<u8>)`?  This would let us write point-free code, e.g. replacing
```rust
    result.and_then(|(_peer, response, _data)| {
        expect_update_chunk_ack(response)
    })?;
```
with
```rust
    result.and_then(expect_update_chunk_ack)?;
```
This works _almost_ everywhere; we'd need to update the signature of `TlvRpc::parse_response` to include those extra arguments.  On the other hand, complex tuple arguments are kinda ugly.  On the third hand, these functions are scoped to the crate, so maybe it doesn't matter that they're specialized for the results of `rpc(..)`?


jgallagher (approved):
This looks fantastic (and much less error-prone), thanks!

---

> Easy: Should we rename sp_response_ext.rs now that it's not actually an extension trait? If so, what should we call it? (my answer is "yes, sp_response_expect.rs")

Yes. `sp_reponse_expect.rs` seems fine to me :+1: 

> Hard: Should the expect_* functions take a tuple of (SocketAddrV6, SpResponse, Vec<u8>)? 

I'm not opposed to making the args special since these aren't `pub` functions. One longstanding low-priority issue is https://github.com/oxidecomputer/management-gateway-service/issues/49: most of the rpc calls ignore the `data` arg entirely, but for completeness probably shouldn't. I'm not sure if passing the data into all the `expect_*`s would make addressing that issue easier (could we do it in the generated fn?) or harder (need to check it separately from the generated fn?). I also don't think it's super urgent to address anyway, so if we think it'd be a big win to pass the tuple in, we could do that now and revisit it whenever we get around to #49.

mkeeter:
Using tuple arguments does make it easier to test `data`!

I split the `expect_` functions into two flavors: those that expect trailing data, and those that do not.  There's now a new macro `expect_data_fn!` which autogenerates code for the former, and error checking in `expect_data!` for the latter.

gateway-sp-comms/src/single_sp.rs
```diff
@@ -423,12 +417,10 @@ impl SingleSp {
...
-        self.rpc(msg).await.and_then(|(_peer, response, _data)| {
-            if persist {
-                expect_component_set_and_persist_active_slot_ack(response)
-            } else {
-                expect_component_set_active_slot_ack(response)
-            }
+        self.rpc(msg).await.and_then(if persist {
+            expect_component_set_and_persist_active_slot_ack
+        } else {
+            expect_component_set_active_slot_ack
```
jgallagher, line 423: Huh, I don't think I'd have thought to use an `if` evaluating to a `fn` like this. Neat!

jgallagher (approved):

mkeeter:
This now closes #49 when merged

PR #110, opened 2023-08-21
Add API for reading a sensor by `SensorId`

mkeeter:
The current plan for metrics is to have IPCC report Hubris sensor IDs as part of its device inventory.  Then, the host will use MGS – with the messages introduced in this PR – to collect data for metrics.

The PR is quite flexible; it adds a new `ReadSensor` message which allows for a bunch of different sub-requests.

(This sidesteps the existing per-MGS-component sensor implementation, which predates the current plan and isn't widely used.  I talked with John about this in advance, and we're in agreement

TODO:
- [x] Bump MGS message version and add exhaustive tests

### DIFF
--- a/gateway-messages/src/lib.rs
+++ b/gateway-messages/src/lib.rs
@@ -63,7 +63,7 @@ pub const MAX_SERIALIZED_SIZE: usize = 1024;
 /// for more detail and discussion.
 pub mod version {
     pub const MIN: u32 = 2;
-    pub const CURRENT: u32 = 7;
+    pub const CURRENT: u32 = 8;
 }
 
 #[derive(
--- a/gateway-messages/src/lib.rs
+++ b/gateway-messages/src/lib.rs
@@ -77,13 +77,17 @@ pub struct Header {
     pub message_id: u32,
 }
 
-#[derive(Debug, Clone, Copy, SerializedSize, Serialize, Deserialize)]
+#[derive(
+    Debug, Clone, Copy, PartialEq, SerializedSize, Serialize, Deserialize,
+)]
 pub struct Message {
     pub header: Header,
     pub kind: MessageKind,
 }
 
-#[derive(Debug, Clone, Copy, Serialize, Deserialize, SerializedSize)]
+#[derive(
+    Debug, Clone, Copy, PartialEq, Serialize, Deserialize, SerializedSize,
+)]
 pub enum MessageKind {
     MgsRequest(MgsRequest),
     MgsResponse(MgsResponse),
@@ -155,7 +159,9 @@ pub enum SensorRequest {
 }
 
 /// Most recent sensor reading, which may be a reading or a value
-#[derive(Debug, Clone, Copy, SerializedSize, Serialize, Deserialize)]
+#[derive(
+    Debug, Clone, Copy, PartialEq, SerializedSize, Serialize, Deserialize,
+)]
 pub struct SensorReading {
     pub value: Result<f32, SensorDataMissing>,
     pub timestamp: u64,
@@ -166,6 +172,7 @@ pub struct SensorReading {
     Debug,
     Clone,
     Copy,
+    PartialEq,
     SerializedSize,
     Serialize,
     Deserialize,
--- a/gateway-messages/src/lib.rs
+++ b/gateway-messages/src/lib.rs
@@ -145,17 +145,24 @@ pub enum SwitchDuration {
 #[derive(
     Debug, Clone, Copy, PartialEq, Eq, SerializedSize, Serialize, Deserialize,
 )]
-pub enum SensorRequest {
-    /// Requests the current timestamp (as a `u64`)
-    CurrentTime,
+pub enum SensorRequestKind {
     /// Requests the most recent reading, which is either a value or error
-    LastReading { id: u32 },
+    LastReading,
     /// Requests the most recent data value
-    LastData { id: u32 },
+    LastData,
     /// Requests the most recent error value
-    LastError { id: u32 },
+    LastError,
     /// Requests the error count for a given sensor
-    ErrorCount { id: u32 },
+    ErrorCount,
+}
+
+/// Sensor readings that we could request from the target by `SensorId`
+#[derive(
+    Debug, Clone, Copy, PartialEq, Eq, SerializedSize, Serialize, Deserialize,
+)]
+pub struct SensorRequest {
+    pub kind: SensorRequestKind,
+    pub id: u32,
 }
 
 /// Most recent sensor reading, which may be a reading or a value
@@ -180,7 +187,6 @@ pub struct SensorReading {
 )]
 #[strum(serialize_all = "snake_case")]
 pub enum SensorResponse {
-    CurrentTime(u64),
     LastReading(SensorReading),
     LastData { value: f32, timestamp: u64 },
     LastError { value: SensorDataMissing, timestamp: u64 },
--- a/gateway-messages/src/lib.rs
+++ b/gateway-messages/src/lib.rs
@@ -77,17 +77,13 @@ pub struct Header {
     pub message_id: u32,
 }
 
-#[derive(
-    Debug, Clone, Copy, SerializedSize, Serialize, Deserialize, PartialEq, Eq,
-)]
+#[derive(Debug, Clone, Copy, SerializedSize, Serialize, Deserialize)]
 pub struct Message {
     pub header: Header,
     pub kind: MessageKind,
 }
 
-#[derive(
-    Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, SerializedSize,
-)]
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, SerializedSize)]
 pub enum MessageKind {
     MgsRequest(MgsRequest),
     MgsResponse(MgsResponse),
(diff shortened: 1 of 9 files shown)


gateway-messages/src/lib.rs
```diff
@@ -141,6 +141,66 @@ pub enum SwitchDuration {
     Forever,
 }
 
+/// Sensor readings that we could request from the target by `SensorId`
+#[derive(
+    Debug, Clone, Copy, PartialEq, Eq, SerializedSize, Serialize, Deserialize,
+)]
+pub enum SensorRequest {
+    /// Requests the current timestamp (as a `u64`)
+    CurrentTime,
```
jgallagher, line 150: This isn't specific to a sensor (or even sensors at all, except that they also return timestamps), right? Should it be a top-level `MgsRequest::CurrentTime` variant? If we did that, we could also consider splitting this enum up into a struct/enum pair to avoid repeating `id`, something like

```rust
pub enum SensorRequestKind { LastReading, LastData, LastError, ErrorCount }

pub struct SensorRequest {
    sensor_id: u32,
    kind: SensorRequestKind,
}
```

gateway-sp-comms/src/single_sp.rs
```diff
@@ -793,6 +796,22 @@ impl SingleSp {
...
+
+    pub async fn read_sensor_value(&self, id: u32) -> Result<SensorReading> {
+        let v = self
+            .rpc(MgsRequest::ReadSensor(SensorRequest::LastReading { id }))
+            .await
+            .and_then(expect_read_sensor)?;
+        match v {
+            SensorResponse::LastReading(r) => Ok(r),
+            // TODO: Is this a bad overloading of
+            // CommunicationError::BadResponseType?
```
jgallagher, line 808: I don't think so. `BadResponseType` should never happen, assuming a well-behaved / correct SP, which also seems true here, right? A `LastReading` request should always get a `LastReading` response.

Should we have other methods for sending the other kinds of `SensorRequest`s?

jgallagher (commented):
LGTM - thanks for writing the (tedious) serialization tests. Couple minor questions.

gateway-messages/src/lib.rs
mkeeter: I like this a lot, done in 6b0afefb7f232caba837f38b3e87c426e99ed561

gateway-sp-comms/src/single_sp.rs
mkeeter: > I don't think so. BadResponseType should never happen, assuming a well-behaved / correct SP, which also seems true here, right? A LastReading request should always get a LastReading response.

Yeah, I was wondering whether we wanted a specialized `SpError::Sensor(SensorError::BadResponseType { ... })`, since the reply / response is nested.  I'm fine keeping it as-is, though, since you don't object!

> Should we have other methods for sending the other kinds of SensorRequests?

Eh, I was planning to add them as needed 🤷🏻‍♂️ 

(Writing the extra request kinds was mostly future-proofing, since now we could hypothetically extend `MgsRequest::ReadSensor` without creating more top-level `MgsRequest` variants)

gateway-messages/src/lib.rs
```diff
@@ -141,6 +141,72 @@ pub enum SwitchDuration {
...
+    PartialEq,
+    SerializedSize,
+    Serialize,
+    Deserialize,
+    strum_macros::IntoStaticStr,
+)]
+#[strum(serialize_all = "snake_case")]
+pub enum SensorResponse {
+    LastReading(SensorReading),
+    LastData { value: f32, timestamp: u64 },
```
rmustacc, line 191: Couple questions here:

1) What is supposed to be teh semantic difference issuing `LastReading` and `LastData`? When would I use one or the other?

2) If there's only a way to get the last reading, does this basically mean that if we're not polling in a hard-real time fashion (that is in the deadline sense of real-time), we're going to lose data in the broader system for any stat that's either discrete or a gauge as opposed to a counter?

3) Out of curiosity, why the choice of normalizing values to floats? I'm guessing that might be what it already was stored as, but it wouldn't have been my first guess / expectation? I assume when we have a discrete sensor like a power good it'll be up to the consumer to know what it represents and translate that into the discrete expected states?

4) As a consumer, it's sometimes helpful to know what the length of time this covers is. For example, we're inserting into Clickhouse and probably telling it that covers a particular range I expect. In particular, for counters, if the data is updating every 5s and we're inserting into Clickhouse every 10s, that makes a difference or vice versa.

5) I assume that the pattern here is that collectors would go back and forth a few times to try to correlate a rolling count in milis with themselves to try and figure out what the relationship between the SP's monotonic time is and the real-world time, which is why you added the alternative way to get the current time. Is that right? I think we'll want to think through what the queuing and retry delays are going to be on how we do this accurately.

@bnaecker can you also take a look at this please to see if this'll give us enough information to actually get things into Clickhouse?

mkeeter: > 1. What is supposed to be teh semantic difference issuing LastReading and LastData? When would I use one or the other?

`LastReading` returns the current state of the sensor, which is either a data value or an error state.  `LastData` returns the last known data value (or a top-level `NoData` error).  I suspect we'll always want to use `LastReading`, but wanted to build in some flexibility.

> 2. If there's only a way to get the last reading, does this basically mean that if we're not polling in a hard-real time fashion (that is in the deadline sense of real-time), we're going to lose data in the broader system for any stat that's either discrete or a gauge as opposed to a counter?

That's currently the case – the Hubris `sensors` task only stores a single data point per sensor.  We could change that (at the cost of more RAM), in which case we'd probably add new variants to the `enum ReadSensor`.

> 3. Out of curiosity, why the choice of normalizing values to floats? I'm guessing that might be what it already was stored as, but it wouldn't have been my first guess / expectation? I assume when we have a discrete sensor like a power good it'll be up to the consumer to know what it represents and translate that into the discrete expected states?

This is also based on how stuff is implemented on the Hubris side.  The actual discussion predate me, but IIRC Cliff had strong opinions about storing all sensor data as floats representing real units (versus, for example, raw bit patterns reported by sensors).

> 4. As a consumer, it's sometimes helpful to know what the length of time this covers is. For example, we're inserting into Clickhouse and probably telling it that covers a particular range I expect. In particular, for counters, if the data is updating every 5s and we're inserting into Clickhouse every 10s, that makes a difference or vice versa.

This is an interesting point.  From the POV of the `sensors` task, readings are simply instantaneous samples (without a notion of range).  It's true that sensors are polled at some frequency, but that's determined by tasks that provide data into `sensors` (and isn't an inherent property of the sensors or an individual reading).

> 5. I assume that the pattern here is that collectors would go back and forth a few times to try to correlate a rolling count in milis with themselves to try and figure out what the relationship between the SP's monotonic time is and the real-world time, which is why you added the alternative way to get the current time. Is that right? I think we'll want to think through what the queuing and retry delays are going to be on how we do this accurately.

I'm not sure what pattern we'll want to implement here!  I'm trying to expose mostly-raw data from the SP, and since the data includes timestamps, I figured we'll want _some_ way to decode them into host time; asking the SP about its current time seems like a useful building block.

bnaecker: @rmustacc From a first glance, there should be no issues getting this data into ClickHouse. I do have a few thoughts, mostly with an eye towards implementing the path to the database.

Something on the host system (MGS, I'm assuming) would want to poll this data "fast enough" to make sure we don't lose our desired temporal resolution. That will probably vary by sensor. MGS or some other component will need to register with `nexus` as a metric producer, at which point an `oximeter` instance will be assigned to collect data from it. That registration includes the collection interval, which is independent of the interval on which the sensors are polled. I'm just pointing out that whatever produces data must (1) be able to make HTTP request to `nexus` and (2) run an HTTP server itself, that is reachable over the underlay network.

The data will need to be slightly reorganized, but nothing major. First `oximeter` supports `f64` measurements, not `f32`. No big deal there, and we can add that if it's important. Next, data needs to be organized into a _target_, the thing we're collecting data about, and a _metric_, the data point being measured. For example, it might make sense to make the thing the sensor is attached to into the target. What the sensor is measuring, say temperature, is the metric. This might be in the weeds implementation-wise, but `Target` and `Metric` are derivable traits in the `oximeter` crate. There are several examples in the `oximeter-producer` crate, `nexus`, `dendrite`, and `crucible`, and I'm always happy to help get more data into ClickHouse.

mkeeter: Thanks for the feedback!  Do you think any of this affects the MGS message types, or can that reorganization be done on the host?

e.g. I'd rather not do `f32 -> f64` conversion on the SP, because I don't think it has hardware support for `f64`'s.

mkeeter:
@bnaecker / @rmustacc How are you feeling after the discussions above?  I'd like to get this merged if possible, but can make further changes if requested.

gateway-messages/src/lib.rs
rmustacc: I discussed this a bit with @bnaecker and I think the organizational part of Targets/Metrics will be what we get from topo so there's nothing more that we need to do at the MGS layer today (assuming that relying on topo is sufficient, which I guess is a bit of a risk).

rmustacc: > > 2. If there's only a way to get the last reading, does this basically mean that if we're not polling in a hard-real time fashion (that is in the deadline sense of real-time), we're going to lose data in the broader system for any stat that's either discrete or a gauge as opposed to a counter?
> 
> That's currently the case – the Hubris `sensors` task only stores a single data point per sensor. We could change that (at the cost of more RAM), in which case we'd probably add new variants to the `enum ReadSensor`.

OK. I think probably the way out of this is to eventually add stats that cover how many times have we seen state X for discrete stats and min/max or some percentile for gauges over time.

> > 4. As a consumer, it's sometimes helpful to know what the length of time this covers is. For example, we're inserting into Clickhouse and probably telling it that covers a particular range I expect. In particular, for counters, if the data is updating every 5s and we're inserting into Clickhouse every 10s, that makes a difference or vice versa.
> 
> This is an interesting point. From the POV of the `sensors` task, readings are simply instantaneous samples (without a notion of range). It's true that sensors are polled at some frequency, but that's determined by tasks that provide data into `sensors` (and isn't an inherent property of the sensors or an individual reading).

I think this becomes most important to think through in the future when we basically have oximeter polling at a different frequency or that it has less real-time fidelity and is trying to poll this source at a different rate. It's probably fine, but just something in my head.

> > 5. I assume that the pattern here is that collectors would go back and forth a few times to try to correlate a rolling count in milis with themselves to try and figure out what the relationship between the SP's monotonic time is and the real-world time, which is why you added the alternative way to get the current time. Is that right? I think we'll want to think through what the queuing and retry delays are going to be on how we do this accurately.
> 
> I'm not sure what pattern we'll want to implement here! I'm trying to expose mostly-raw data from the SP, and since the data includes timestamps, I figured we'll want _some_ way to decode them into host time; asking the SP about its current time seems like a useful building block.

OK, makes sense.

rmustacc:
@mkeeter I'm fine at my end.

PR #116, opened 2023-08-31
Add ReadRot message for CMPA / CFPA

mkeeter:
This PR adds a new `ReadRot` top-level message, for all of our RoT reading needs (right now, just CMPA / CFPA).

The CMPA / CFPA results are a fixed size of 512 bytes, but I'm still putting them in the trailing data buffer to avoid blowing up `SpResponse::MAX_SIZE`.

This PR should be left open until the Hubris side is also done, in case we discover the need for further changes (e.g. new error variants)


gateway-messages/tests/versioning/v8.rs
```diff
@@ -70,15 +70,15 @@ fn host_request() {
         let request =
             MgsRequest::ReadSensor(SensorRequest { kind, id: 0x1234 });
         let mut expected = vec![
-            38, // SpResponse::ReadSensor
+            38, // MgsRequest::ReadSensor
```
mkeeter, line 73: Changes in this file are typo fixes, unrelated to the rest of the PR

gateway-sp-comms/src/single_sp.rs
```diff
@@ -814,6 +816,24 @@ impl SingleSp {
             }),
         }
     }
+
+    pub async fn read_rot_cmpa(&self) -> Result<[u8; 512]> {
```
jgallagher, line 820: One nit and one question:

* Could we add a constant (or possibly two, for naming clarity? I'm ambivalent about this bit) for the CMPA / CFPA size? As written I think clients of this crate will also need to create some `[u8; 512]`s, and constants would probably be friendlier.
* How is a client (MGS, or possibly the client of MGS) supposed to interpret the returned blob? Will hubtools take this as input for some kind of "display what we know about this" or "compare against a known list" operation?

labbott: I think the interpretation depends on what problems we are actually trying to solve. We can use this with the verify operation in lpc55_sign to validate a particular Hubris image will work with a particular CMPA/CFPA combination. 

The constant is the LPC55 flash page size. We have that defined in a driver but I don't know the best place to put it for other not-LPC55 software. We could also cheat and use the `block_size` call to get the same answer.

gateway-messages/src/lib.rs
```diff
@@ -207,6 +222,27 @@ pub enum SensorDataMissing {
...
+#[derive(
+    Debug, Clone, Copy, PartialEq, Eq, SerializedSize, Serialize, Deserialize,
+)]
+pub enum CfpaPage {
+    /// Currently active page
+    Active,
+    /// Page that will become active on reset
+    Pending,
+    // We could add variants for ping/pong/scratch pages here if needed
+}
```
labbott, line 244: I think we should be more explicit to cover all cases
```
enum CfpaPage {
     Active, // HIghest version between ping/pong
     Alternate, // Lower version between ping/pong
     Scratch, // Data to be applied on next reboot
}
```
Given the difficulty in getting data out of the RoT, I'd like to guarantee we can actually get all possible data we want and then leave interpretation of that elsewhere. My fear is that the RoT tries to calculate what will happen with "pending", something goes wrong and then we're stuck with confusing data when it doesn't get applied.

gateway-sp-comms/src/single_sp.rs
mkeeter: Sure, added `ROT_PAGE_SIZE` in db88706aa5d6b434ed428712445798b89b8f1ef5.

I agree with @labbott that we'd probably pass the pages into `hubtools` directly, since that's the existing API.

gateway-messages/src/lib.rs
mkeeter: Sure, I can do that.  Do you want `Active` / `Alternate`, or should we get even more manual and report `Ping` / `Pong` (leaving the host to interpret which one is active)?

mkeeter: Okay, switched to `Active` / `Inactive` / `Scratch` in 39455d696e1367377bec6e8c102cf034a837ac1f

PR #121, opened 2023-09-08
Add SprotProtocolError::Desynchronized

andrewjstone:
This is an error type that will be plumbed up from the RoT. This change should be merged after #116, as it bumps the version to `v10`, where that #116 bumps it to `v9`.


mkeeter (approved):

andrewjstone:
After talking with @mkeeter, I changed the version here to v9. #116 will instead become v10.

PR #172, opened 2023-12-15
Use slog-error-chain for error formatting and logging

jgallagher:
Corrects usages of `#[from]` and `#[source]` with embedded error strings as described in [slog-error-chain](https://github.com/oxidecomputer/slog-error-chain?tab=readme-ov-file#aside-embedding-source-error-strings), and uses that crate to add `slog::{KV,Value}` impls to all our error types.

Goodbye

```
Error: Error response from SP: the image caboose does not contain 'NONE': the image caboose does not contain 'NONE'
```

Hello

```
Error: Error response from SP: the image caboose does not contain 'NONE'
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -510,7 +510,6 @@ dependencies = [
  "slog-async",
  "slog-term",
  "termios",
- "thiserror",
  "tokio",
  "tokio-stream",
  "tokio-util",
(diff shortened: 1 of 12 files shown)


andrewjstone (approved):
🎉 🎉 🎉

PR #209, opened 2024-03-21
Update stage0

### LINKED ISSUE #208: Support update of RoT bootloader
The RoT bootloader needs to be updatable in the field.

lzrd:
Support update of RoT bootloader.

Includes always calculating FWID and signature check at RoT boot time.
Messages for SP and RoT Hubris updates are backward compatible.

See https://github.com/oxidecomputer/sprot-e2e for test scripts that exercise update and rollback of Hubris and Bootleby on SP and RoT.

A new RoTBootInfo mesage is plumbed that includes info on the four RoT flash banks.

The ImageError enum gives detailed information on invalid RoT flash bank contents.

This is a draft PR because it needs to be coordinated with hubris and omicron changes so as not to break people's development environments.

Closes #208

### DIFF
--- a/faux-mgs/Cargo.toml
+++ b/faux-mgs/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "faux-mgs"
-version = "0.1.0"
+version = "0.1.1"
 edition = "2021"
 license = "MPL-2.0"
 
--- a/faux-mgs/src/main.rs
+++ b/faux-mgs/src/main.rs
@@ -19,6 +19,7 @@ use gateway_messages::ComponentAction;
 use gateway_messages::IgnitionCommand;
 use gateway_messages::LedComponentAction;
 use gateway_messages::PowerState;
+use gateway_messages::RotBootInfo;
 use gateway_messages::SpComponent;
 use gateway_messages::StartupOptions;
 use gateway_messages::UpdateId;
@@ -178,6 +179,16 @@ enum Command {
     },
 
     /// Get or set the active slot of a component (e.g., `host-boot-flash`).
+    ///
+    /// Except for component "stage0", setting the active slot can be
+    /// viewed as an atomic operation.
+    ///
+    /// Setting "stage0" slot 1 as the active slot initiates a copy from
+    /// slot 1 to slot 0 if the contents of slot 1 still match those seen
+    /// at last RoT reset and the contents are properly signed.
+    ///
+    /// Power failures during the copy can disable the RoT. Only one stage0
+    /// update should be in process in a rack at any time.
     ComponentActiveSlot {
         #[clap(value_parser = parse_sp_component)]
         component: SpComponent,
@@ -373,6 +384,14 @@ enum Command {
 
     /// Reads the lock status of any VPD in the system
     VpdLockStatus,
+
+    /// Read the RoT's boot-time information.
+    RotBootInfo {
+        /// Return highest version of RotBootInfo less then or equal to our
+        /// highest known version.
+        #[clap(long, short, default_value_t = RotBootInfo::HIGHEST_KNOWN_VERSION)]
+        version: u8,
+    },
 }
 
 #[derive(Subcommand, Debug, Clone)]
@@ -827,9 +846,12 @@ async fn run_command(
                     ));
                     lines.push(format!("hubris version: {:?}", state.version));
                     lines.push(format!("power state: {:?}", state.power_state));
-
-                    // TODO: pretty print RoT state?
-                    lines.push(format!("RoT state: {:?}", state.rot));
+                    match state.rot {
+                        Ok(rot) => {
+                            lines.push(format!("rot: Ok({})", rot.display()))
+                        }
+                        Err(err) => lines.push(format!("rot: Err({})", err)),
+                    }
                     Ok(Output::Lines(lines))
                 }
                 VersionedSpState::V2(state) => {
@@ -857,10 +879,50 @@ async fn run_command(
                             .join(":")
                     ));
                     lines.push(format!("power state: {:?}", state.power_state));
-                    // TODO: pretty print RoT state?
-                    lines.push(format!("RoT state: {:?}", state.rot));
+                    match state.rot {
+                        Ok(rot) => lines.push(format!("rot: Ok({})", rot)),
+                        Err(err) => lines.push(format!("rot: Err({})", err)),
+                    }
                     Ok(Output::Lines(lines))
                 }
+                VersionedSpState::V3(state) => {
+                    lines.push(format!(
+                        "hubris archive: {}",
+                        hex::encode(state.hubris_archive_id)
+                    ));
+
+                    lines.push(format!(
+                        "serial number: {}",
+                        zero_padded_to_str(state.serial_number)
+                    ));
+                    lines.push(format!(
+                        "model: {}",
+                        zero_padded_to_str(state.model)
+                    ));
+                    lines.push(format!("revision: {}", state.revision));
+                    lines.push(format!(
+                        "base MAC address: {}",
+                        state
+                            .base_mac_address
+                            .iter()
+                            .map(|b| format!("{b:02x}"))
+                            .collect::<Vec<_>>()
+                            .join(":")
+                    ));
+                    lines.push(format!("power state: {:?}", state.power_state));
+                    Ok(Output::Lines(lines))
+                }
+            }
+        }
+        Command::RotBootInfo { version } => {
+            let rot_state = sp.rot_state(version).await?;
+            info!(log, "{rot_state:x?}");
+            if json {
+                Ok(Output::Json(serde_json::to_value(rot_state).unwrap()))
+            } else {
+                let mut lines = Vec::new();
+                lines.push(format!("{}", rot_state.display()));
+                Ok(Output::Lines(lines))
             }
         }
         Command::Ignition { target } => {
--- a/gateway-messages/src/lib.rs
+++ b/gateway-messages/src/lib.rs
@@ -66,7 +66,7 @@ pub const ROT_PAGE_SIZE: usize = 512;
 /// for more detail and discussion.
 pub mod version {
     pub const MIN: u32 = 2;
-    pub const CURRENT: u32 = 12;
+    pub const CURRENT: u32 = 13;
 
     /// MGS protocol version in which SP watchdog messages were added
     pub const WATCHDOG_VERSION: u32 = 12;
@@ -129,7 +129,7 @@ pub enum BadRequestReason {
     DeserializationError,
 }
 
-/// Image slot name for SwitchDefaultImage
+/// Image slot name for SwitchDefaultImage on component ROT
 #[derive(
     Debug, Clone, Copy, PartialEq, Eq, SerializedSize, Serialize, Deserialize,
 )]
--- a/gateway-messages/src/mgs_to_sp.rs
+++ b/gateway-messages/src/mgs_to_sp.rs
@@ -204,6 +204,11 @@ pub enum MgsRequest {
     ComponentWatchdogSupported {
         component: SpComponent,
     },
+
+    /// Read RoT boot state at the highest version not to exceed specified version.
+    VersionedRotBootInfo {
+        version: u8,
+    },
 }
 
 #[derive(
--- a/gateway-messages/src/sp_to_mgs.rs
+++ b/gateway-messages/src/sp_to_mgs.rs
@@ -131,6 +131,9 @@ pub enum SpResponse {
 
     DisableComponentWatchdogAck,
     ComponentWatchdogSupportedAck,
+
+    SpStateV3(SpStateV3),
+    RotBootInfo(RotBootInfo),
 }
 
 /// Identifier for one of of an SP's KSZ8463 management-network-facing ports.
@@ -165,6 +168,38 @@ pub struct DiscoverResponse {
 pub struct ImageVersion {
     pub epoch: u32,
     pub version: u32,
+}
+
+#[derive(
+    Debug, Copy, Clone, PartialEq, Eq, Deserialize, Serialize, SerializedSize,
+)]
+pub enum ImageError {
+    /// Image has not been sanity checked (internal use)
+    Unchecked = 1,
+    /// First page of image is erased.
+    FirstPageErased,
+    /// Some pages in the image are erased.
+    PartiallyProgrammed,
+    /// The NXP image offset + length caused a wrapping add.
+    InvalidLength,
+    /// The header flash page is erased.
+    HeaderNotProgrammed,
+    /// A bootloader image is too short.
+    BootloaderTooSmall,
+    /// A required ImageHeader is missing.
+    BadMagic,
+    /// The image size in ImageHeader is unreasonable.
+    HeaderImageSize,
+    /// total_image_length in ImageHeader is not properly aligned.
+    UnalignedLength,
+    /// Some NXP image types are not supported.
+    UnsupportedType,
+    /// Wrong format reset vector.
+    ResetVectorNotThumb2,
+    /// Reset vector points outside of image execution range.
+    ResetVector,
+    /// Signature check on image failed.
+    Signature,
 }
 
 /// This is quasi-deprecated in that it will only be returned by SPs with images
@@ -203,11 +238,72 @@ pub struct SpStateV2 {
 }
 
 #[derive(
+    Debug, Clone, Copy, PartialEq, Eq, SerializedSize, Serialize, Deserialize,
+)]
+pub struct SpStateV3 {
+    pub hubris_archive_id: [u8; 8],
+    // Serial and revision are only 11 bytes in practice; we have plenty of room
+    // so we'll leave the fields wider in case we grow it in the future. The
+    // values are 0-padded.
+    pub serial_number: [u8; 32],
+    pub model: [u8; 32],
+    pub revision: u32,
+    pub base_mac_address: [u8; 6],
+    pub power_state: PowerState,
+}
+
+type Digest256 = [u8; 32];
+
+trait HexDisplayableArray {
+    // fn display(&self) -> fmt::Display;
+    fn display(&self) -> HexStringDisplay;
+}
+
+impl HexDisplayableArray for Digest256 {
+    fn display(&self) -> HexStringDisplay<'_> {
+        HexStringDisplay(&self[..])
+    }
+}
+
+struct HexStringDisplay<'a>(&'a [u8]);
+
+impl<'a> fmt::Display for HexStringDisplay<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "\"")?;
+        for b in self.0.iter() {
+            write!(f, "{:02x}", b)?;
+        }
+        write!(f, "\"")?;
+        Ok(())
+    }
+}
+
+#[derive(
     Debug, Clone, Copy, PartialEq, Eq, Deserialize, Serialize, SerializedSize,
 )]
 pub struct RotImageDetails {
-    pub digest: [u8; 32],
+    pub digest: Digest256,
     pub version: ImageVersion,
+}
+
+impl RotImageDetails {
+    pub fn display(&self) -> RotImageDetailsDisplay<'_> {
+        RotImageDetailsDisplay(self)
+    }
+}
+
+/// This class exists for faux-mgs to nicely display the Firmware ID (FWID).
+#[derive(Clone, Debug)]
+#[must_use = "this struct does nothing unless displayed"]
+pub struct RotImageDetailsDisplay<'a>(pub &'a RotImageDetails);
+
+impl<'a> fmt::Display for RotImageDetailsDisplay<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let s = self.0;
+        writeln!(f, "digest: {}, ", s.digest.display())?;
+        writeln!(f, "version: {:?}, ", s.version)?;
+        Ok(())
+    }
 }
 
 /// The boot time details dumped by Stage0 into Hubris on the RoT
@@ -218,6 +314,47 @@ pub struct RotBootState {
     pub active: RotSlotId,
     pub slot_a: Option<RotImageDetails>,
     pub slot_b: Option<RotImageDetails>,
+}
+
+impl RotBootState {
+    pub fn display(&self) -> RotBootStateDisplay<'_> {
+        RotBootStateDisplay(self)
+    }
+}
+
+/// This class exists for faux-mgs to nicely display the Firmware ID (FWID).
+#[derive(Clone, Debug)]
+#[must_use = "this struct does nothing unless displayed"]
+pub struct RotBootStateDisplay<'a>(pub &'a RotBootState);
+
+impl<'a> fmt::Display for RotBootStateDisplay<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let s = self.0;
+        writeln!(f, "active: {:?}, ", s.active)?;
+        write!(f, "slot_a: ")?;
+        match s.slot_a {
+            Some(details) => {
+                writeln!(
+                    f,
+                    "Some(RotImageDetails{{ {} }}),",
+                    &RotImageDetailsDisplay(&details)
+                )?;
+            }
+            None => writeln!(f, "None,")?,
+        };
+        write!(f, "slot_b: ")?;
+        match s.slot_b {
+            Some(details) => {
+                writeln!(
+                    f,
+                    "Some(RotImageDetails{{ {} }}),",
+                    &RotImageDetailsDisplay(&details)
+                )?;
+            }
+            None => writeln!(f, "None,")?,
+        };
+        Ok(())
+    }
 }
 
 #[derive(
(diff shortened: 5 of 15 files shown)


faux-mgs/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "faux-mgs"
-version = "0.1.0"
+version = "0.2.0"
```
labbott, line 3: We haven't been bumping the version it looks like, I'm not opposed though.

gateway-sp-comms/src/single_sp/update.rs
```diff
@@ -261,53 +262,127 @@ fn read_auxi_check_from_tlvc(data: &[u8]) -> Result<[u8; 32], UpdateError> {
...
+                            break;
+                        }
+                    }
+                    rot_image
+                }
+                _ => return Err(UpdateError::InvalidSlotIdForOperation),
+            }
+        }
+        _ => return Err(UpdateError::InvalidComponent),
+    };
```
labbott, line 382: Can we follow a similar pattern to what we have in embootleby? https://github.com/oxidecomputer/embootleby/blob/main/src/main.rs#L193-L239 We try the hubtools format first and if that doesn't work fall back to manually reading the zip archive. I think that will be cleaner to remove in the future.

gateway-sp-comms/src/sp_response_expect.rs
```diff
@@ -348,6 +351,24 @@ mod tests {
...
+        let v = expect_sp_state((
+            dummy_addr(),
+            SpResponse::SpStateV3(state),
+            vec![],
+        ));
+        let Ok(r) = v else {
+            panic!("mismatched value {v:?}");
+        };
+        assert_eq!(r, VersionedSpState::V3(state));
     }
```
labbott, line 372: I'm pretty sure if we're adding new entries here we need a protocol bump and corresponding tests (c.f. https://github.com/oxidecomputer/management-gateway-service/commit/e727e960d31f62afe6077fdfc405e0ac2379f6b2)

gateway-messages/src/sp_to_mgs.rs
```diff
@@ -199,6 +233,21 @@ pub struct SpStateV2 {
...
+    pub hubris_archive_id: [u8; 8],
+    // Serial and revision are only 11 bytes in practice; we have plenty of room
+    // so we'll leave the fields wider in case we grow it in the future. The
+    // values are 0-padded.
+    pub serial_number: [u8; 32],
+    pub model: [u8; 32],
+    pub revision: u32,
+    pub base_mac_address: [u8; 6],
+    pub power_state: PowerState,
+}
```
labbott, line 249: What's the reasoning behind decoupling RoT state from the SP? To allow changes in the future?

gateway-messages/src/sp_to_mgs.rs
```diff
@@ -570,6 +835,9 @@ pub enum SpError {
     Update(UpdateError),
     Sensor(SensorError),
     Vpd(VpdError),
+    BlockOutOfOrder,
+    InvalidSlotIdForOperation,
+    InvalidComponent,
```
jgallagher, line 840: Why are these variants added both here and to `UpdateError`?

gateway-messages/src/sp_to_mgs.rs
```diff
@@ -788,6 +1072,27 @@ impl fmt::Display for UpdateError {
             Self::MissingHandoffData => {
                 write!(f, "boot data not handed off to hubris kernel")
             }
+            Self::BlockOutOfOrder => {
+                write!(f, "bootloader blocks delivered out of order")
```
jgallagher, line 1076: Tiny nit - the `UpdateError` variant here could (at least in theory) come from something other than the bootloader - should we omit `"bootloader"` from the output?

(Same comment on `VersionNotSupported` below, where it might not come from RoT boot info?)

gateway-sp-comms/src/single_sp/update.rs
```diff
@@ -257,57 +258,137 @@ fn read_auxi_check_from_tlvc(data: &[u8]) -> Result<[u8; 32], UpdateError> {
     })
 }
 
+/// Isolate extraction of bootleby from old-format archives.
+/// TODO: Purge old-format archives and delete this code.
```
jgallagher, line 262: Is this `TODO` to be addressed on this PR, or at some point in the future? (I don't know off hand when we switched away from `old-format`)

gateway-sp-comms/src/single_sp/update.rs
```diff
@@ -257,57 +258,137 @@ fn read_auxi_check_from_tlvc(data: &[u8]) -> Result<[u8; 32], UpdateError> {
...
+    let mut rot_image = vec![];
+    for i in 0..archive.len() {
+        let mut file = match archive.by_index(i) {
+            Ok(file) => file,
+            Err(e) => {
+                error!(
+                    log,
+                    "did not find bootleby.bin in old-style archive: {e}"
+                );
+                return Err(UpdateError::InvalidArchive);
```
jgallagher, line 284: `zip`'s API is a little weird, but I don't think this error is correct, right? E.g., you could have an archive with 10 files, file 7 has an I/O error, and `bootleby.bin` is file 8. We should probably log the error but `continue`, I think.

gateway-sp-comms/src/single_sp/update.rs
```diff
@@ -257,57 +258,137 @@ fn read_auxi_check_from_tlvc(data: &[u8]) -> Result<[u8; 32], UpdateError> {
...
+        if file.name() == "bootleby.bin" {
+            if file.read_to_end(&mut rot_image).is_err() {
+                error!(log, "invalid archive");
+                return Err(UpdateError::InvalidArchive);
+            }
+            debug!(log, "using bootleby.bin from old-style archive");
+            break;
+        }
+    }
+    Ok(rot_image)
```
jgallagher, line 296: This doesn't seem right if we don't find a `bootleby.bin` - it will return an empty vec, right? Could we change the success case in the loop to `return Ok(rot_image)` instead of `break`, then return an error here instead?

gateway-sp-comms/src/single_sp/update.rs
```diff
@@ -257,57 +258,137 @@ fn read_auxi_check_from_tlvc(data: &[u8]) -> Result<[u8; 32], UpdateError> {
...
+                    // be able to test update and rollback using the oldest
+                    // releases that may be in customers' racks or spares pool.
+                    bootleby_from_old_style_archive(image, log)
+                        // Report the original Hubtools error if
+                        // this second chance did not work.
+                        .map_err(|_| hubtool_error))?
+
+            // TODO: Even though the RoT will protect itself, put
+            // pre-flash checks here for BORD, Bootloader vs Hubris,
+            // and signature validity.
```
jgallagher, line 385: Do you want to do this TODO as part of this PR or a followup?

faux-mgs/Cargo.toml
lzrd: Changed to 0.1.1

gateway-sp-comms/src/single_sp/update.rs
lzrd: I made a change to that effect. Thanks!

gateway-messages/src/sp_to_mgs.rs
lzrd: Going down this path and allowing for cases where either the SP or the RoT had the new version, I could see an exponential growth. Not so bad when the exponent is only 2, but why not decouple?

lzrd: Although our model is that you update RoT first, then SP, we can't really promise  that that will always be the case.

faux-mgs/src/main.rs
```diff
@@ -131,6 +131,14 @@ enum Command {
     /// Ask SP for its current state.
     State,
 
+    /// Ask RoT via SP for its current boot-time information represented by
+    /// a particular struct.
+    RotBootInfo {
+        /// RotStateV2 is the default verion
+        #[clap(long, short, default_value = "3")]
+        version: u8,
```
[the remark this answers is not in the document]
lzrd: Yes, this is a backward/forward compatibility issue to allow updates when there is version skew between SP and RoT. We can deprecate messages using rollback protection (to be implemented) and ensuring that sufficient old images have been eradicated from the fleet. Barring that, we may have a stepping stone like guideline saying that you can go from A to D, but you have to go through C and D to get there.

gateway-messages/src/sp_to_mgs.rs
```diff
@@ -255,6 +304,222 @@ pub struct RotStateV2 {
...
+    pub slot_a_status: Result<(), ImageError>,
+    /// Slot B status at last RoT reset
+    pub slot_b_status: Result<(), ImageError>,
+    /// Stage0 (bootloader) status at last RoT reset
+    pub stage0_status: Result<(), ImageError>,
+    /// Stage0Next status at last RoT reset
+    pub stage0next_status: Result<(), ImageError>,
+}
+
+impl fmt::Display for RotState {
```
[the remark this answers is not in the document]
lzrd: This is here to serve faux-mgs display of return messages. These were missing and the humans had to look at numeric dumps.

I'll take a look though.

gateway-messages/src/sp_to_mgs.rs
```diff
@@ -255,6 +304,222 @@ pub struct RotStateV2 {
...
+        writeln!(f, "stage0next_status: {:?},", self.stage0next_status)?;
+
+        write!(f, "}}")?;
+        Ok(())
+    }
+}
+
+/// While the `rot_state` API is deprecated, the RotState structures remain.
+/// They are fetched with `rot_boot_info` API for `RotState`, and
+/// `versioned_rot_boot_info` to get the latest or a particular version.
```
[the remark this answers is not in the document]
lzrd: I should explain more clearly.

The rot_boot_info sprot message was only used by the SP when composing its reply to `faux-mgs ... state`. The version of the SP response that contains rot_boot_info needs to be deprecated. The new `versioned_rot_boot_info` message is separate and plumbed through to MGS.

gateway-messages/src/sp_to_mgs.rs
lzrd: you are correct

gateway-sp-comms/src/single_sp/update.rs
lzrd: Do we want to proceed using an archive that has a file in it that can't be parsed?

gateway-sp-comms/src/single_sp/update.rs
lzrd: Good catch.

gateway-sp-comms/src/sp_response_expect.rs
lzrd: I added tests to gateway_messages/tests/versioning/v12.rs that include SpStateV3 and other new symbols.

gateway-messages/src/sp_to_mgs.rs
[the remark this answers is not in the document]
lzrd: I did as you suggested. This affects the non-JSON output to make it easier for a person to work with the FWIDs.
```
$ for x in 1 2 3; do echo === $x; faux-mgs-dev --log-level=CRITICAL rot-boot-info --version $x; done |& tee out
=== 1
 RotBootInfo { V1(RotState { rot_updates: RotUpdateDetails { boot_state: RotBootState { active: B, slot_a: Some(RotImageDetails{ digest: "dc60d4fef774c17404f8417fd2849ca85e11e8878cf3974eea2c071649bca7c6", version: ImageVersion { epoch: 0, version: 0 },  }), slot_b: Some(RotImageDetails{ digest: "e9c138fdeb0215412d9cd077c8303a413121d4cdc12a2f3c70ed2ca8b50bae9c", version: ImageVersion { epoch: 0, version: 0 },  }),  } } }
)}

=== 2
 RotBootInfo { V2(RotStateV2 {
active: B, persistent_boot_preference: B, pending_persistent_boot_preference: None, transient_boot_preference: None, slot_a_sha3_256_digest: Some("dc60d4fef774c17404f8417fd2849ca85e11e8878cf3974eea2c071649bca7c6"), slot_b_sha3_256_digest: Some("e9c138fdeb0215412d9cd077c8303a413121d4cdc12a2f3c70ed2ca8b50bae9c"), })}

=== 3
 RotBootInfo { V3(RotStateV3 {
active: B, persistent_boot_preference: B, pending_persistent_boot_preference: None, transient_boot_preference: None, slot_a_fwid: Fwid::Sha3_256("dc60d4fef774c17404f8417fd2849ca85e11e8878cf3974eea2c071649bca7c6"), slot_b_fwid: Fwid::Sha3_256("e9c138fdeb0215412d9cd077c8303a413121d4cdc12a2f3c70ed2ca8b50bae9c"), stage0_fwid: Fwid::Sha3_256("e24ceb8532f3ee4de15ed393243bad351058adbf4b9fde5e1863302d97c42c60"), stage0next_fwid: Fwid::Sha3_256("62ebe27addfd99e6f6e2ba75302bbb96f8a2d4b2d01b270cbff2ee4e4a2c1280"), slot_a_status: Ok(()), slot_b_status: Ok(()), stage0_status: Ok(()), stage0next_status: Ok(()) })}
```

gateway-sp-comms/src/single_sp/update.rs
lzrd: I clarified the text: "When old-format archives are eliminated from customer racks and spares inventory, then this code can be removed."

gateway-sp-comms/src/single_sp/update.rs
lzrd: That should be a follow-up. [Issue 221](https://github.com/oxidecomputer/management-gateway-service/issues/221)

gateway-messages/src/sp_to_mgs.rs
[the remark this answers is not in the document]
lzrd: I elaborated in the comment and opened  issue #222 for rollback protection.

faux-mgs/src/main.rs
```diff
@@ -373,6 +385,13 @@ enum Command {
 
     /// Reads the lock status of any VPD in the system
     VpdLockStatus,
+
+    /// Read the RoT's boot-time information.
+    RotBootInfo {
+        /// Return highest version of RotBootInfo less then or equal to given
+        #[clap(long, short, default_value = "3")]
```
jgallagher, line 392: I don't love the `"3"` here because I'm worried it'll be easy to forget to update; if we add a new boot info version, `3` will still be valid but won't be the default we'd actually want, right?

I'm not sure how best to fix this. Maybe we could add a `RotBootInfo::latest_version()` function and use that as the default value here? It has the same problem in that it would just return `3` today, but I think it'd be much less likely to be overlooked if someone is adding a variant to `RotBotInfo`.

faux-mgs/src/main.rs
```diff
@@ -857,11 +879,55 @@ async fn run_command(
                             .join(":")
                     ));
                     lines.push(format!("power state: {:?}", state.power_state));
-                    // TODO: pretty print RoT state?
-                    lines.push(format!("RoT state: {:?}", state.rot));
+                    match state.rot {
+                        Ok(rotstate) => {
+                            lines.push(format!(
+                                "{}",
+                                &RotStateV2Display(&rotstate)
```
jgallagher, line 886: Can we put `.display()` helper methods on the relevant types so a caller doesn't have to construct this by hand? E.g.,

```rust
lines.push(format!("{}", rotstate.display()));
```

gateway-messages/src/lib.rs
```diff
@@ -138,6 +138,25 @@ pub enum RotSlotId {
...
+
+#[derive(
+    Debug, Clone, Copy, PartialEq, Eq, SerializedSize, Serialize, Deserialize,
+)]
+pub enum ComponentSlot {
+    /// Hubris flash slot
+    Rot(RotSlotId),
+    /// Bootloader flash slot
+    Stage0(Stage0SlotId),
+}
```
jgallagher, line 158: Are these two enums used anywhere?

gateway-messages/src/mgs_to_sp.rs
```diff
@@ -204,6 +204,11 @@ pub enum MgsRequest {
     ComponentWatchdogSupported {
         component: SpComponent,
     },
+
+    /// Read RoT boot state at the highest version not to exceed specified version.
```
jgallagher, line 208: Just a question about the wording here: I understand this to mean that if MGS requests `version: 5` and the RoT supports up to `version: 8`, the RoT should send back a version 5 state struct. What if MGS requests `version: 5` and the RoT only supports up to `version: 3`?

gateway-messages/src/sp_to_mgs.rs
```diff
@@ -210,6 +260,20 @@ pub struct RotImageDetails {
...
 
+/// This class exists for faux-mgs to nicely display the Firmware ID (FWID).
+#[derive(Clone, Debug)]
+#[must_use = "this struct does nothing unless displayed"]
+pub struct RotImageDetailsDisplay<'a>(pub &'a RotImageDetails);
+
+impl<'a> fmt::Display for RotImageDetailsDisplay<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let s = self.0;
+        write!(f, "digest: {}, ", &Digest256Display(&s.digest))?;
```
jgallagher, line 271: (Same note about adding an `s.digest.display()` helper; I'll stop leaving this comment, but I think we should add them for all these display adapters.)

gateway-messages/src/sp_to_mgs.rs
jgallagher: Bump to include this question in this review

gateway-messages/tests/versioning/v12.rs
```diff
@@ -12,14 +12,21 @@
...
+use gateway_messages::RotSlotId;
+use gateway_messages::RotStateV3;
 use gateway_messages::RotWatchdogError;
 use gateway_messages::SerializedSize;
 use gateway_messages::SpComponent;
 use gateway_messages::SpError;
 use gateway_messages::SpResponse;
 use gateway_messages::WatchdogError;
 
+use gateway_messages::SpStateV3;
```
jgallagher, line 28: With apologies for the messiness of this: v12 has already landed on main (with #216), so these tests need to move to a `v13.rs` module, and the protocol version needs to be bumped to 13.

gateway-sp-comms/src/single_sp/update.rs
```diff
@@ -257,57 +258,142 @@ fn read_auxi_check_from_tlvc(data: &[u8]) -> Result<[u8; 32], UpdateError> {
...
+                        Ok(_) => {
+                            debug!(
+                                log,
+                                "using bootleby.bin from old-style archive"
+                            );
+                            return Ok(rot_image);
+                        }
+                        Err(err) => {
+                            error!(log, "cannot access bootleby.bin from zip file index {i}: {err}");
+                            break;
```
jgallagher, line 291: Nit - we should probably return an error here; breaking means we'll return `UpdateError::ImageNotFound`, but that isn't the same as "there was a bootleby.bin but we failed to read it".

gateway-messages/tests/versioning/v12.rs
lzrd: Yes. There was hope that they would land on the same release. I will change that.

faux-mgs/src/main.rs
lzrd: Since versions could be skewed, we want greatest common understood version. So not really "latest_version" but more like RotBootInfo::limit_highest_common_version() or something.

faux-mgs/src/main.rs
lzrd: This is only for human convenience in working with FWIDs. I could take this stuff out and add it later if it is still useful. It's a bunch of code with limited value since I have since worked out how to deal with  this in a jq script.

gateway-messages/src/lib.rs
lzrd: There is no higher level code updating stage0 right now. There is code in Hubris. I need to check on its use.

faux-mgs/src/main.rs
jgallagher: I'm not sure that's relevant to my comment? :sweat_smile: We still have all this code, and it would be nicer to have `.display()` methods than needing to construct the adapters by hand.

faux-mgs/src/main.rs
jgallagher: I'm not sure I follow that. How would faux-mgs know what the greatest common understood version is?

lzrd: It knows what it knows, it doesn't know what the RoT knows. But, once faux-mgs tells the RoT: "Give me the highest version you  know not greater than this version" the greatest common version is known.

gateway-messages/src/mgs_to_sp.rs
lzrd: Rot returns version 3 then. It doesn't know anything higher.

faux-mgs/src/main.rs
jgallagher: Based on https://github.com/oxidecomputer/management-gateway-service/pull/209#discussion_r1575316155, shouldn't this always be "the highest version faux-mgs knows about"? Which would make `latest_version()` make sense?

gateway-messages/tests/versioning/v12.rs
jgallagher: FWIW, I don't think this is relevant to what release they land in. Since we can install off-release builds, we've been bumping the version in any PR that makes additions to the protocol.

faux-mgs/src/main.rs
lzrd: I added a pub const next to the RotBootInfo definition and now have faux-mgs using that constant.

gateway-messages/src/lib.rs
lzrd: Nope. Thanks for spotting this cruft.

faux-mgs/src/main.rs
```diff
@@ -373,6 +387,14 @@ enum Command {
 
     /// Reads the lock status of any VPD in the system
     VpdLockStatus,
+
+    /// Read the RoT's boot-time information.
+    RotBootInfo {
+        /// Return highest version of RotBootInfo less then or equal to our
+        /// highest known version.
+        #[clap(long, short, default_value = formatcp!("{}", RotBootInfo::HIGHEST_KNOWN_VERSION))]
```
jgallagher, line 395: Nit - we don't need const formatting; this can be

```suggestion
        #[clap(long, short, default_value_t = RotBootInfo::HIGHEST_KNOWN_VERSION)]
```

as long as the type matches. (It looks like `HIGHEST_KNOWN_VERSION` is a `u32` - it should be a `u8`, right?)

gateway-messages/tests/versioning/v12.rs
lzrd: done

faux-mgs/src/main.rs
lzrd: right. done

gateway-sp-comms/src/single_sp/update.rs
lzrd: I changed this to return UpdateError::InvalidArchive

gateway-messages/tests/versioning/v13.rs
```diff
@@ -0,0 +1,123 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! The tests in this module check that the serialized form of messages from MGS
+//! protocol version 12 have not changed.
```
jgallagher, line 6:
```suggestion
//! protocol version 13 have not changed.
```

gateway-messages/tests/versioning/v13.rs
```diff
@@ -0,0 +1,123 @@
...
+//!
+//! If a test in this module fails, _do not change the test_! This means you
+//! have changed, deleted, or reordered an existing message type or enum
+//! variant, and you should revert that change. This will remain true until we
+//! bump the `version::MIN` to a value higher than 11, at which point these
+//! tests can be removed as we will stop supporting v11.
```
jgallagher, line 12:
```suggestion
//! bump the `version::MIN` to a value higher than 13, at which point these
//! tests can be removed as we will stop supporting v13.
```

(or feel free to reword this to omit the number, so the next person that grabs this comment block for a new file doesn't have to update it)

gateway-messages/tests/versioning/v13.rs
```diff
@@ -0,0 +1,123 @@
...
+use gateway_messages::ImageError;
+use gateway_messages::MgsRequest;
+use gateway_messages::RotBootInfo;
+use gateway_messages::RotSlotId;
+use gateway_messages::RotStateV3;
+use gateway_messages::SerializedSize;
+use gateway_messages::SpResponse;
+use gateway_messages::SpStateV3;
+
+#[test]
```
jgallagher, line 25: With continued apologies for the tediousness of this, can you also add a test for the new `ImageError` enum, the new `UpdateError` variants, and possibly the new `SpError` variants (although I'm not sure those should exist - [I asked about them in the first review](https://github.com/oxidecomputer/management-gateway-service/pull/209#discussion_r1559803409) but it may have gotten lost in the github comment collapsing).

gateway-messages/src/sp_to_mgs.rs
lzrd: Neglected to delete from SpError during development apparently. Thanks.

gateway-messages/tests/versioning/v13.rs
lzrd: Done

gateway-messages/tests/versioning/v13.rs
lzrd: Reworded for future selves.

gateway-messages/src/sp_to_mgs.rs
lzrd: I think I have dealt with all of these.

jgallagher (approved):

gateway-messages/src/sp_to_mgs.rs
```diff
@@ -167,6 +170,38 @@ pub struct ImageVersion {
     pub version: u32,
 }
 
+#[derive(
+    Debug, Copy, Clone, PartialEq, Eq, Deserialize, Serialize, SerializedSize,
+)]
+pub enum ImageError {
+    /// Image has not been sanity checked (internal use)
+    Unchecked = 1,
```
mkeeter, line 178: Do we ever care about the in-memory representation of this `enum`?  If not, I'd recommend removing the ` = 1`, because it _suggests_ that we care about the in-memory representation.

Note that the ` = 1` **does not** change how this object will be serialized.  For example:
```rust
#[derive(serde::Serialize)]
enum Foo {
    One = 1,
    Two,
}

fn main() {
    let b = bincode::serialize(&Foo::One).unwrap();
    println!("{b:?}") // prints [0, 0, 0, 0]
}
```

lzrd:
You're right, we don't care about the in-memory representation.

PR #215, opened 2024-04-01
Update USDT dependency

bnaecker:


bnaecker:
Just an outdated dep I noticed while doing unrelated work on the dependency graph in Omicron!

jgallagher (approved):
Thanks!

Issue #141, opened 2023-10-10
SP addressing has TOCTOU race


davepacheco:
Clients of MGS address service processors by type and slot number (e.g., `/sp/sled/10`), which I think corresponds to cubby numbers (or maybe more precisely: backplane cable numbers).  But these can presumably change over time (e.g., if an operator moves sleds between cubbies) and I don't see a way for a client to avoid accidentally sending a command to the wrong SP.

Concretely, if a client is looking to turn off a particular sled (that they know by its serial number), they might:

1. for each slot, use `GET /sp/sled/$slot` to see which sled is in which slot and store this
2. pick the slot that reported having the sled it wants to turn off
3. use `POST /sp/sled/$slot/power-state` to move the sled to A2

But if an operator has moved the sled between cubbies between steps 1 and 3, we might turn off some other sled!

---

@jgallagher and I talked about a few ways to deal with this.  I think in order to avoid the time-of-check-to-time-of-use race, the SP has to be involved in verifying that the request is intended for it.  MGS could include in its messages to the SP some identifier that the SP can reject if it doesn't match its own identifier.  This identifier could maybe be the serial number, or it could be some made-up identifier that MGS has previously assigned to the SP (which would presumably need to be re-sent by MGS if the SP restarts).

PR #216, opened 2024-04-02
Add SP reset watchdog

mkeeter:
Per https://github.com/oxidecomputer/hubris/issues/1695, flashing a broken image onto the SP currently requires physical intervention; if we can't talk to the SP over MGS, we have to pull the sled, attach a debugger, and reflash it.

This PR adds message types to enable and disable a watchdog, then uses those message before resetting the SP (which happens as part of the update process); if the SP resets into a broken image, the watchdog will switch back to the known-working image.

This is the MGS counterpart to https://github.com/oxidecomputer/hubris/pull/1707

### DIFF
--- a/faux-mgs/src/main.rs
+++ b/faux-mgs/src/main.rs
@@ -325,7 +325,7 @@ enum Command {
 
     /// Instruct the SP to reset.
     Reset {
-        /// Disables the SP slot watchdog
+        /// Reset without the automatic safety rollback watchdog
         #[clap(long)]
         disable_watchdog: bool,
     },
@@ -337,7 +337,7 @@ enum Command {
     ResetComponent {
         #[clap(value_parser = parse_sp_component)]
         component: SpComponent,
-        /// Disables any watchdogs associated with this component reset
+        /// Reset without the automatic safety rollback watchdog (if applicable)
         #[clap(long)]
         disable_watchdog: bool,
     },
--- a/faux-mgs/src/main.rs
+++ b/faux-mgs/src/main.rs
@@ -324,7 +324,11 @@ enum Command {
     },
 
     /// Instruct the SP to reset.
-    Reset,
+    Reset {
+        /// Disables the SP slot watchdog
+        #[clap(long)]
+        disable_watchdog: bool,
+    },
 
     /// Reset a component.
     ///
@@ -333,6 +337,9 @@ enum Command {
     ResetComponent {
         #[clap(value_parser = parse_sp_component)]
         component: SpComponent,
+        /// Disables any watchdogs associated with this component reset
+        #[clap(long)]
+        disable_watchdog: bool,
     },
 
     /// Controls the system LED
@@ -1139,10 +1146,14 @@ async fn run_command(
                 }
             }
         }
-        Command::Reset => {
+        Command::Reset { disable_watchdog } => {
             sp.reset_component_prepare(SpComponent::SP_ITSELF).await?;
             info!(log, "SP is prepared to reset");
-            sp.reset_component_trigger(SpComponent::SP_ITSELF).await?;
+            sp.reset_component_trigger(
+                SpComponent::SP_ITSELF,
+                disable_watchdog,
+            )
+            .await?;
             info!(log, "SP reset complete");
             if json {
                 Ok(Output::Json(json!({ "ack": "reset" })))
@@ -1151,10 +1162,10 @@ async fn run_command(
             }
         }
 
-        Command::ResetComponent { component } => {
+        Command::ResetComponent { component, disable_watchdog } => {
             sp.reset_component_prepare(component).await?;
             info!(log, "SP is prepared to reset component {component}",);
-            sp.reset_component_trigger(component).await?;
+            sp.reset_component_trigger(component, disable_watchdog).await?;
             info!(log, "SP reset component {component} complete");
             if json {
                 Ok(Output::Json(json!({ "ack": "reset" })))
--- a/gateway-messages/src/mgs_to_sp.rs
+++ b/gateway-messages/src/mgs_to_sp.rs
@@ -188,6 +188,13 @@ pub enum MgsRequest {
     /// communicate with the SP enough to call this function, then it's safe to
     /// disable the watchdog (because the new image is working).
     DisableSpSlotWatchdog,
+
+    /// Checks whether the SP watchdog is supported
+    ///
+    /// It may be unsupported if the SP is running an older version of MGS (in
+    /// which case this message will fail to decode), or because there's a
+    /// debugger connected to the SP (preventing the RoT from controlling it).
+    SpSlotWatchdogSupported,
 }
 
 #[derive(
--- a/gateway-messages/src/mgs_to_sp.rs
+++ b/gateway-messages/src/mgs_to_sp.rs
@@ -174,27 +174,36 @@ pub enum MgsRequest {
     /// The values are serialized in the trailer of the packet
     VpdLockState,
 
-    /// Reset the SP after enabling the RoT watchdog
+    /// Reset the given component after enabling its fail-safe watchdog
     ///
-    /// This function is part of the SP reset sequence, and will return an error
-    /// if there is not a completed update in the SP.
-    ResetWithWatchdog {
+    /// Right now, the only valid target for this command is `SP_ITSELF`.
+    /// The command will return an error if no update is staged, because
+    /// rollback is meaningless in that situation.
+    ResetComponentTriggerWithWatchdog {
+        component: SpComponent,
         time_ms: u32,
     },
 
-    /// Disable the RoT watchdog
+    /// Disable the rollback watchdog for the given component
     ///
-    /// This function should be called after `ResetWithWatchdog`; if we can
-    /// communicate with the SP enough to call this function, then it's safe to
-    /// disable the watchdog (because the new image is working).
-    DisableSpSlotWatchdog,
+    /// Right now, the only valid component for this command is `SP_ITSELF`.
+    ///
+    /// This should be called after `ResetComponentTriggerWithWatchdog`; if we
+    /// can communicate with the SP enough to call this function, then it's safe
+    /// to disable the watchdog (because the new image is working).
+    DisableComponentWatchdog {
+        component: SpComponent,
+    },
 
-    /// Checks whether the SP watchdog is supported
+    /// Checks whether a rollback watchdog is supported
     ///
     /// It may be unsupported if the SP is running an older version of MGS (in
     /// which case this message will fail to decode), or because there's a
-    /// debugger connected to the SP (preventing the RoT from controlling it).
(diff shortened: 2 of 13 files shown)


gateway-messages/tests/versioning/v11.rs
```diff
@@ -36,7 +36,7 @@ fn host_request() {
 
 #[test]
 fn vpd_protocol_errors() {
-    let mut out = [0; MgsRequest::MAX_SIZE];
+    let mut out = [0; SpResponse::MAX_SIZE];
```
mkeeter, line 39: This is a drive-by fix, because I noticed we're encoding an `SpResponse` below (not an `MgsRequest`)

gateway-sp-comms/src/single_sp.rs
```diff
@@ -99,6 +100,15 @@ const DISCOVERY_INTERVAL_IDLE: Duration = Duration::from_secs(60);
...
 const TLV_RPC_TOTAL_ITEMS_DOS_LIMIT: u32 = 1024;
 
+// We allow our client to specify the max RPC attempts and the
+// per-attempt timeout; however, it's very easy to set a timeout that is
+// too low for the "reset the SP" request, especially if the SP being
+// reset is a sidecar (which means it won't be able to respond until it
+// brings the management network back online). We will override the max
+// attempt count for only that message to ensure we give SPs ample time
+// to reset.
+const SP_RESET_TIME_ALLOWED: Duration = Duration::from_secs(30);
```
andrewjstone, line 110: In the interest of not making things worse and always accidentally resetting to the old version, maybe we should make this even longer. What about 60, or even 120 seconds?

It only makes sense to extend the timeout if we keep trying periodically and don't have to wait the whole timeout though.

andrewjstone (approved):

gateway-sp-comms/src/single_sp.rs
```diff
@@ -740,10 +750,67 @@ impl SingleSp {
...
     ) -> Result<()> {
+        // MGS protocol version in which SP watchdog messages were added
+        const MGS_WATCHDOG_VERSION: u32 = 12;
+
+        // If the SP has an update pending, then use the watchdog reset
+        let mut use_watchdog = component == SpComponent::SP_ITSELF
+            && !disable_watchdog
+            && matches!(
+                self.update_status(component).await?,
+                UpdateStatus::Complete(..)
```
jgallagher, line 770: Is this check strictly required for correctness, or would it be okay to use the watchdog even for non-update resets? If it is required, we might have a similar problem as #141 in that this is inherently racy, and some other client could reset the SP in between this check and the subsequent requests this method makes. If it's not required, I'd be tempted to remove this check, because (a) I think the vast majority of resets are probably to apply an update status and (b) it removes something that looks like a TOCTOU race.

gateway-sp-comms/src/single_sp.rs
```diff
@@ -740,10 +750,67 @@ impl SingleSp {
...
+    /// If `disable_watchdog` is `true`, then any watchdogs associated with the
+    /// reset are disabled.  Otherwise, watchdogs are enabled opportunistically
+    /// (depending on component and MGS protocol version).
     pub async fn reset_component_trigger(
         &self,
         component: SpComponent,
+        disable_watchdog: bool,
     ) -> Result<()> {
+        // MGS protocol version in which SP watchdog messages were added
+        const MGS_WATCHDOG_VERSION: u32 = 12;
```
jgallagher, line 763: Tiny nit - I wonder if it would make sense to put this inside the `gateway_messages::version` module, so any future explicit version checks start to collect similar constants in one place?

gateway-messages/src/mgs_to_sp.rs
```diff
@@ -173,6 +173,28 @@ pub enum MgsRequest {
     /// Dump information about the lock state of the VPD (Vital Product Data)
     /// The values are serialized in the trailer of the packet
     VpdLockState,
+
+    /// Reset the SP after enabling the RoT watchdog
+    ///
+    /// This function is part of the SP reset sequence, and will return an error
+    /// if there is not a completed update in the SP.
+    ResetWithWatchdog {
```
jgallagher, line 181: Do you think there's any chance we'll use a similar watchdog patterns for any other components? We originally had `ResetPrepare/ResetTrigger`, then later added `ResetComponentPrepare/ResetComponentTrigger` (with support for component `SP_ITSELF`), and I'm wondering if we should skip to that second step this time around.

gateway-messages/src/sp_to_mgs.rs
```diff
@@ -1033,3 +1040,27 @@ impl fmt::Display for VpdError {
...
+    /// Could not control the SP over SWD
+    SpCtrl,
+    /// There is not a complete SP update in place
+    NoCompletedUpdate,
+}
+
+impl fmt::Display for WatchdogError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::NoCompletedUpdate => {
```
jgallagher, line 1060: Oh, this might answer my earlier question about the potential TOCTOU (I clearly haven't looked at the hubris side of this yet!). Is this right?

* If we try to reset with watchdog to an SP that does not have a completed update, it will return this error?
* If so, (a) the check for a completed update _is_ required and (b) the TOCTOU is solved by this on the SP side (since if one client checks for an update, then gets paused while another client resets the SP, then resumes, the reset will fail with this error)?

gateway-messages/tests/versioning/v12.rs
```diff
@@ -0,0 +1,86 @@
...
+fn rot_watchdog_error() {
+    let mut out = [0; RotError::MAX_SIZE];
+
+    let err = WatchdogError::SpCtrl;
+    // using a match to force exhaustive checking here
+    let serialized = match err {
+        WatchdogError::SpCtrl => [5, 0],
+        WatchdogError::NoCompletedUpdate => [5, 1],
+    };
+    // TODO do we need both SpError::Watchdog and RotError::Watchdog?
```
jgallagher, line 83: I don't follow this TODO given both `watchdog_error()` and `rot_watchdog_error()` exist - is it leftover?

faux-mgs/src/main.rs
```diff
@@ -333,6 +337,9 @@ enum Command {
     ResetComponent {
         #[clap(value_parser = parse_sp_component)]
         component: SpComponent,
+        /// Disables any watchdogs associated with this component reset
```
jgallagher, line 340: I'm tempted to say this reinforces my concern that the watchdog message should carry a component along with it - this help message seems a little misleading if it only applies to the SP.

I'm also not sure how to word this, exactly; "disable the watchdog" is a thing that internally we _want_ to do after enabling it. Maybe something like "Reset without the automatic rollback safety watchdog"?

gateway-sp-comms/src/single_sp.rs
mkeeter: This timeout didn't actually change – I just pulled it into a global constant so that I could use it in more than one place.

gateway-sp-comms/src/single_sp.rs
mkeeter: Sure, done in 747d750d97ce3beadd1073d7d62c2f093fe1bc5c

gateway-messages/tests/versioning/v12.rs
mkeeter: I was displeased by `RotError` (in MGS) containing a `Watchdog` variant with the full `WatchdogError` type, given that the RoT shouldn't know about `WatchdogError::NoCompletedUpdate`.

This was necessary in the first place is because we need to report watchdog errors in `SprotError` (through the SP-RoT SPI link), and there's an `impl From<SprotError> for RotError` (for... reasons?), so we needed to represent watchdog errors in `RotError` as well.

I've changed this in e82f1076f402deeb85b9713b60c356da2f0c6ffe to split off the RoT-related watchdog errors into their own enum, so `RotError::Watchdog` now only knows about a relevant subset of errors.

gateway-messages/src/mgs_to_sp.rs
mkeeter: Yes, I switched over to component-flavored messages.

faux-mgs/src/main.rs
mkeeter: Rephrased in c5960b41297263f7cff89c378e52b403ca49c8ae

(and yes, there's now a component associated with the reset-with-watchdog message)

gateway-messages/src/sp_to_mgs.rs
mkeeter: Yes, your interpretation is correct.  There shouldn't be a TOCTOU issue because the SP will reject the `ResetComponentTriggerWithWatchdog` command if there's no update staged, so the check earlier is just for more ergonomic error handling.

gateway-sp-comms/src/single_sp.rs
mkeeter: See https://github.com/oxidecomputer/management-gateway-service/pull/216#discussion_r1566297077 ; this check is for ergonomics, but the SP also does the same check atomically as part of applying the `ResetComponentTriggerWithWatchdog` request.

gateway-sp-comms/src/single_sp.rs
```diff
@@ -740,10 +751,66 @@ impl SingleSp {
...
+    /// If `disable_watchdog` is `true`, then any watchdogs associated with the
+    /// reset are disabled.  Otherwise, watchdogs are enabled opportunistically
+    /// (depending on component and MGS protocol version).
     pub async fn reset_component_trigger(
         &self,
         component: SpComponent,
+        disable_watchdog: bool,
     ) -> Result<()> {
+        // If the SP has an update pending, then use the watchdog reset
+        let mut use_watchdog = component == SpComponent::SP_ITSELF
```
jgallagher, line 764: Now that we're taking a `component` in the message, should we drop this check for `SP_ITSELF` here? Relatedly, do you think `component_watchdog_supported` should return a `bool` instead of just an ack? I think if it did, that would make "is the watchdog supported for this component" is an SP-level decision instead of an MGS-level decision.

jgallagher: (I guess the "clearing a watchdog" logic would similarly have to change, although maybe that is also just removing the `SP_ITSELF` check and relying on `use_watchdog`?)

jgallagher (approved):
With apologies for the protracted back and forth - just one final comment about the transition to taking a `component`. If you make changes related to that that you want more review for I'm happy to do so, but if you're content (with or without changes to address that) I'm equally happy for this to go ahead and land.

gateway-sp-comms/src/single_sp.rs
mkeeter: I think this is reasonable; done in 78b7356edbea64fe30e96b1451806037f4a80f71 (along with some other cleanup)

PR #230, opened 2024-05-08
Allow faux-mgs to specify 'stage0' in read-component-caboose.

lzrd:
Fix an omission in faux-mgs to allow read-component caboose to address RoT stage0.


labbott (approved):
LGTM

PR #254, opened 2024-07-23
Add tech port unlocking

mkeeter:
This PR adds tech port unlocking as described in [RFD 492](https://rfd.shared.oxide.computer/rfd/492).

It should be read alongside the corresponding Hubris PR (https://github.com/oxidecomputer/hubris/pull/1832)

There are a bunch of changes, which I'll summarize below.

## Adding the `Sender` type
```rust
pub struct Sender {
    /// Address of the sender
    pub addr: SocketAddrV6,
    /// SP port on which the packet was received
    pub port: SpPort,
    /// VLAN tag associated with the received packet
    ///
    /// This is often one-to-one equivalent to `port`; however, on Sidecar,
    /// multiple VLANs are mapped to `SpPort::One`.
    pub vid: u16,
}
```
This type is passed as an argument to `SpImpl` functions that care about the sender.  For functions which do not care about the sender, the pre-existing `sender: SocketAddrV6, port: SpPort,` arguments have been removed.

## Allow `ComponentAction` to return data
Added `enum ComponentActionResponse` as the return type for `SpImpl::component_action`, so that component actions can return data.

For backwards compatibility, `ComponentActionResponse::Ack` is translated into `SpResponse::ComponentActionAck`.

## Added new types
`MonorailComponentAction`, `MonorailComponentActionResponse`, `UnlockChallenge`, `UnlockResponse`; all of the pieces to perform tech port unlocking.

## Add `monorail lock/unlock` subcommand to `faux-mgs`
This is where the locking and unlocking actually gets done!

### DIFF
--- a/faux-mgs/src/main.rs
+++ b/faux-mgs/src/main.rs
@@ -38,6 +38,7 @@ use gateway_sp_comms::SwitchPortConfig;
 use gateway_sp_comms::VersionedSpState;
 use gateway_sp_comms::MGS_PORT;
 use serde_json::json;
+use slog::debug;
 use slog::info;
 use slog::o;
 use slog::warn;
@@ -421,6 +422,10 @@ enum MonorailCommand {
     Unlock {
         /// How long to unlock for
         time_sec: u32,
+
+        /// Path to private key for SSH signing challenge
+        #[clap(long)]
+        key: Option<PathBuf>,
     },
 
     /// Lock the technician port
@@ -1318,8 +1323,8 @@ async fn run_command(
                     )
                     .await?
                 }
-                MonorailCommand::Unlock { time_sec } => {
-                    monorail_unlock(&log, &sp, time_sec).await?;
+                MonorailCommand::Unlock { time_sec, key } => {
+                    monorail_unlock(&log, &sp, time_sec, key).await?;
                 }
             }
             if json {
@@ -1435,6 +1440,7 @@ async fn monorail_unlock(
     log: &Logger,
     sp: &SingleSp,
     time_sec: u32,
+    key: Option<PathBuf>,
 ) -> Result<()> {
     let r = sp
         .component_action_with_response(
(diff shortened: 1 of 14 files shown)


mkeeter:
Before:
```
Used:
  flash:   0xa8100 (65%)
  ram:     0x4f800 (62%)
  sram1:   0x4000 (12%)
  sram2:   extern region (jefe, dump_agent)
  sram3:   extern region (jefe, dump_agent)
  sram4:   extern region (jefe, dump_agent)
  bank2:   extern region (update_server)
```

After:
```
Used:
  flash:   0xb8100 (71%)
  ram:     0x67800 (80%)
  sram1:   0x4000 (12%)
  sram2:   extern region (jefe, dump_agent)
  sram3:   extern region (jefe, dump_agent)
  sram4:   extern region (jefe, dump_agent)
  bank2:   extern region (update_server)
```

Here's the per-task size diff:
```diff
98,99c92,93
<                      ram     5384    6256    (fixed)
< jefe                 flash   8620    8704    16384
---
>                      ram     5552    6256    (fixed)
> jefe                 flash   8608    8704    16384
102a97,98
> rng_driver           flash   6244    6272    8192
>                      ram     452     512     512
107,108c103,104
< net                  flash   79048   81920   131072
<                      ram     44256   65536   65536
---
> net                  flash   83500   98304   131072
>                      ram     79104   131072  131072
110,112c106,108
< control_plane_agent  flash   105992  106496  131072
<                      ram     28476   32768   32768
< sprot                flash   45328   45344   65536
---
> control_plane_agent  flash   136520  136576  140672
>                      ram     35352   65536   65536
> sprot                flash   45336   45344   65536
114,119c110,115
< udpecho              flash   8952    8960    16384
<                      ram     4428    8192    8192
< udpbroadcast         flash   8460    8480    16384
<                      ram     4488    8192    8192
< monorail             flash   80148   80384   262144
<                      ram     9108    16384   16384
---
> udpecho              flash   9008    9024    16384
>                      ram     4460    8192    8192
> udpbroadcast         flash   8508    8512    16384
>                      ram     4520    8192    8192
> monorail             flash   81728   81920   262144
>                      ram     9236    16384   16384
130,131c126,127
< transceivers         flash   39236   39264   65536
<                      ram     11044   16384   16384
---
> transceivers         flash   39352   39360   65536
>                      ram     11076   16384   16384
140c136
< validate             flash   8792    8800    16384
---
> validate             flash   8796    8800    16384
146,147c142,143
< dump_agent           flash   18204   18432   32768
<                      ram     12916   16384   16384
---
> dump_agent           flash   18248   18432   32768
>                      ram     12948   16384   16384
```

faux-mgs/src/main.rs
```diff
@@ -1387,6 +1436,96 @@ async fn run_command(
...
+    info!(log, "received challenge {challenge:?}");
+
+    let response = match challenge {
+        UnlockChallenge::EcdsaSha2Nistp256(data) => {
+            let Some(priv_path) = key else {
+                bail!("need --key for ECDSA challenge");
+            };
+
+            debug!(log, "loading private key from {priv_path:?}");
+            let priv_key = ssh_key::PrivateKey::read_openssh_file(&priv_path)
```
jgallagher, line 1469: I don't think this works with a yubikey, right? There's no way to read the private key, and it can only be interacted with via ssh-agent.

I'd be fine landing this as-is and then doing the ssh-agent work as a followup, but just to think out loud about that work for a sec:

* Should we talk directly to ssh-agent via something like https://crates.io/crates/ssh-agent-client-rs, or should we shell out to `ssh-keygen`? I'm tempted to lean toward the former, although that crate in particular might need some minor work [to support sha256 in its API](https://docs.rs/ssh-agent-client-rs/0.9.1/src/ssh_agent_client_rs/codec.rs.html#91-95).
* Once we add ssh-agent support, do we need to keep raw private key file support? I'd lean toward "no" but could be convinced otherwise since it wouldn't be hard to keep around.

faux-mgs/src/main.rs
```diff
@@ -1387,6 +1436,96 @@ async fn run_command(
...
+            // They are a pair of `mpint` values, per
+            // https://datatracker.ietf.org/doc/html/rfc4251
+            //
+            // Each one is either 32 bytes or 33 bytes with a leading zero, so
+            // we'll awkwardly allow for both cases.
+            let sig_bytes = signed.signature_bytes();
+            let mut signature = [0u8; 64];
+            let start = match sig_bytes[3] {
+                32 => 4,
+                33 => 5,
```
jgallagher, line 1506: Should we sanity check that `sig_bytes[4] == 0` in this case?

faux-mgs/src/main.rs
```diff
@@ -1387,6 +1436,96 @@ async fn run_command(
...
+                i => bail!("invalid length {i}"),
+            };
+            signature[0..32].copy_from_slice(&sig_bytes[start..][..32]);
+            let tail = sig_bytes.len() - 32;
```
jgallagher, line 1510: Do we need to check the length of `sig_bytes` to ensure that this doesn't overlap with `[start..][..32]`?

gateway-sp-comms/src/single_sp.rs
```diff
@@ -841,6 +843,11 @@ impl SingleSp {
         // If we are resetting the SP itself, then reset trigger should
         // retry until we get back an error indicating the
         // SP wasn't expecting a reset trigger (because it has reset!).
+        //
+        // On Sidecar, we will instead get a message back indicating that the
+        // management network is locked (if we're updating the SP from a
+        // temporarily-unlocked tech port).
```
jgallagher, line 849: Ooh nice catch; I definitely wouldn't have thought of this.

gateway-messages/src/sp_impl.rs
```diff
@@ -697,27 +635,24 @@ fn handle_message_impl<H: SpHandler>(
...
 fn handle_mgs_response<H: SpHandler>(
-    sender: SocketAddrV6,
-    port: SpPort,
+    sender: Sender,
     message_id: u32,
     handler: &mut H,
     kind: MgsResponse,
     leftover: &[u8],
 ) {
+    // TODO check sender here?
```
jgallagher, line 644: I think yes? `faux-mgs` will never send a `MessageKind::MgsResponse`, but an arbitrary program could, and they should presumably be filtered out immediately if they're coming from a locked tech port.

mkeeter:
Closing in favor of #254

PR #259, opened 2024-07-29
Tech port unlocking (simple version)

mkeeter:
Equivalent to #254, but without any policies besides `UnlockChallenge::Trivial`

### DIFF
--- a/gateway-messages/src/mgs_to_sp.rs
+++ b/gateway-messages/src/mgs_to_sp.rs
@@ -315,21 +315,21 @@ pub enum MonorailComponentAction {
     Lock,
 }
 
-/// Actions for the Monorail switch
+/// Challenge provided to the SP when someone wants to unlock it
 #[derive(
     Copy, Clone, Serialize, SerializedSize, Deserialize, PartialEq, Eq, Debug,
 )]
 pub enum UnlockChallenge {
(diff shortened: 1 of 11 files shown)


jgallagher (approved):

gateway-messages/src/mgs_to_sp.rs
```diff
@@ -292,6 +293,45 @@ pub enum LedComponentAction {
...
+    Lock,
+}
+
+/// Actions for the Monorail switch
+#[derive(
+    Copy, Clone, Serialize, SerializedSize, Deserialize, PartialEq, Eq, Debug,
+)]
+pub enum UnlockChallenge {
+    /// Unlock given an [UnlockResponse::Trivial]
+    Trivial,
```
lzrd, line 324: It would be good to see a nonce here. I know it doesn't affect security, but it does detect challenges that are out of sync with responses, a class of bugs that we could iron out now.

lzrd (approved):

gateway-messages/src/mgs_to_sp.rs
mkeeter: Good idea, added a timestamp in 56e4bccdc158 (deliberately calling it `timestamp` instead of `nonce` so no one starts thinking it's cryptographic)

PR #271, opened 2024-08-23
Make `SingleSp::set_power_state` idempotent

### LINKED ISSUE #270: `SingleSp::set_power_state` should probably be a bit more idempotent
At present, if one attempts to set a SP to a power state that it's already in, one receives a 503 Service Unavailable. For instance, attempting to update Sled 14 on `madrid` failed when `wicketd` attempted to put it in A2, because the system was already in A2:

```log
[         mupdate] [sled 14 00:00:00]   Running  7/19) Setting host power state to A2
[         mupdate] [sled 14 00:00:00]    Failed  7/19) Setting host power state to A2: after 2.38ms: updating power state failed
[         mupdate] [sled 14 00:00:00]             Caused by:
[         mupdate] [sled 14 00:00:00]             - Error Response: status: 503 Service Unavailable; headers: {"content-type": "application/json", "x-request-id": "f64e6790-ccfa-43e0-a629-f1ee71d32c5b", "content-length": "233", "date": "Sun, 28 Dec 1986 00:05:40 GMT"}; value: Error { error_code: Some("SpCommunicationFailed"), message: "error communicating with SP SpIdentifier { typ: Sled, slot: 14 }: Error response from SP: power state error (code 1))", request_id: "f64e6790-ccfa-43e0-a629-f1ee71d32c5b" }
[         mupdate] [sled 14 00:00:00]  Terminal Execution failed
```

This appears to be because the `gimlet-seq` task in Hubris returns an error for `A2` -> `A2` transitions. See this match:
https://github.com/oxidecomputer/hubris/blob/2f24dd4419634ca95742de901d18e52ff14b912b/drv/gimlet-seq-server/src/main.rs#L696-L918

`gateway-sp-comms`' `SingleSp::set_power_state` method just tries to perform the RPC, bubbling up the error from the SP without checking whether it's already in the desired state:

https://github.com/oxidecomputer/management-gateway-service/blob/7518095686028d80b8125de62ba7d179cbc2a683/gateway-sp-comms/src/single_sp.rs#L652-L657

aaaaaaand finally, over in Omicron, the MGS HTTP API just bubbles up whatever error `SingleSp::set_power_state` returns: 
https://github.com/oxidecomputer/omicron/blob/d96ea7ca8945b8ad78a53fd083850ea39789e5f0/gateway/src/http_entrypoints.rs#L623-L637

We should probably make this method succeed when the host is already in the desired state, instead. Although we *could* make the sequencer task in Hubris responsible for this, it seems to me that it might make more sense to do it here in MGS?

hawkw:
Presently, the `SingleSp::set_power_state` method can fail if the system
is already in the requested power state, which is a bit of a bummer.
This is because the Hubris sequencer task returns an `IllegalTransition`
error when the requested and current power states are the same.

This commit changes it to first check the power state, and not ask the
SP to do anything else if it's already in the requested state.

Closes #270

### DIFF
--- a/gateway-sp-comms/src/single_sp.rs
+++ b/gateway-sp-comms/src/single_sp.rs
@@ -651,6 +651,17 @@ impl SingleSp {
 
     /// Set the current power state.
     pub async fn set_power_state(&self, power_state: PowerState) -> Result<()> {
+        // First, let's make sure the system isn't already *in* the desired power
+        // state --- the SP will return an error if we try to set the power
+        // state to the state it's already in.
+        if power_state == self.power_state().await? {
+            trace!(
+                &self.log,
+                "already in the desired power state, doing nothing";
+                "power_state" => ?power_state,
+            );
+            return Ok(());
+        }
         self.rpc(MgsRequest::SetPowerState(power_state))
             .await
             .and_then(expect_set_power_state_ack)


hawkw:
Alternatively, we could just make the Hubris sequencer handle this case, but @jclulow suggested that the current behavior made sense at the level of the sequencer...I don't have a strong opinion on this. I'd be happy to change whatever.

jclulow:
I like the property that the lowest level thing will report that you asked it to do something non-sensical, and that it's up to the higher level orchestration software to get both hands around the state of the thing.  If the sequencer task also ignores a transition from A2 to A2, then it will be harder to detect unexpected things happening upstack in the future.

gateway-sp-comms/src/single_sp.rs
```diff
@@ -651,6 +651,17 @@ impl SingleSp {
 
     /// Set the current power state.
     pub async fn set_power_state(&self, power_state: PowerState) -> Result<()> {
+        // First, let's make sure the system isn't already *in* the desired power
+        // state --- the SP will return an error if we try to set the power
+        // state to the state it's already in.
+        if power_state == self.power_state().await? {
```
jgallagher, line 657: Do we care that there's a TOCTOU issue here (i.e., some other request flips the power state to A2 in between this check and line 665)? It looks like the SP gives back a `PowerStateError(u32)` where one of the u32 values presumably maps to `SeqError::IllegalTransition` (at least for gimlet - I didn't check sidecar).

A couple options:

* Merge this as is, maybe noting the TOCTOU and that we probably (?) don't care about it.
* Add a more specific error variant that the SP can return for this; then this method could issue `SetPowerState` directly and choose to ignore if we get the "you're already in this state" error?

jgallagher: We just had an error performing a mupdate in the factory that I think means we _do_ care about this TOCTOU:

Screenshot from 2024-09-17 13-28-27

I believe what happened here was:

* Sled was in A2
* wicket instructed MGS to change the sled's power state to A0
* MGS sends that request to the SP, then a couple things happen:
  * The SP does receive this request, and transitions the sled to A0
  * MGS did not receive a response within 2 seconds
* MGS's current config is to repeat requests every 2 seconds until success or timeout, so it resent the request to go to A0 after 2 seconds
* The SP rejected this request because the sled was in A0

This is based on a couple things:

* The sled was in A0 when we checked by hand after this failure
* The error is at _exactly_ 2 seconds, which matches how long MGS waits before resending its request

The change in this PR would not have prevented this error, I think? Because the initial check would have passed (the sled _was_ in A2), but then the "operation not idempotent" failure happened later. I think the way MGS does retries is relying on individual requests/packets being idempotent, so I think we probably want something like my second suggestion above (teach MGS how to tell if the error is "I'm already in this state" and then treat that as success).

hawkw: IMO, the potential TOCTOU issue was an argument in favor of making this change inside the SP instead...but, I think we could alternatively make MGS handle that specific error, instead.

jgallagher: I guess we have 3 places where we could choose to do that:

* SP sequencer task
* SP control-plane-agent task
* MGS

I don't have any particularly strong preference. A weak preference is that right now, the specific error code control-plane-agent sends to MGS in this case is opaque and not good to match on, so we'd need SP changes _anyway_. Maybe we could put swallowing this error into control-plane-agent?

jclulow: If we're going to swallow it in the SP I think we should at least include a bit in the result that one could use to reason about whether the action had any effect or was for some reason a no-op.

jgallagher: We can do that, but I'm not sure it makes sense to in the situation outlined above? If the sled is in A2 and MGS requests a transition to A0, we'll usually get back "success and I made the change", but if that response gets lost or MGS times out or whatever and resends the request, we could get back "success and it was already in A0", even though the only reason it was already in A0 was that we actually did receive and handle the first attempt of this same request.

rmustacc: This came up again today while doing another mupdate, my suggestion is a little different in that it says that the SP should include the **current state** when it generates this error which allows the caller to figure out what to do and how to proceed with as much information as possible.

hawkw: I like @rmustacc's suggestion. It avoids the potential TOCTOU issues of trying to paper over this in MGS, but also doesn't have the Hubris sequencer completely hiding the error from upstack software. Also, it's a fairly straightforward change that doesn't require complex logic in the sequencer. I'll give that a go.

hawkw:
This change is obsoleted by the Hubris changes from oxidecomputer/hubris#2064 and oxidecomputer/hubris#2065, and the MGS changes from #390

PR #274, opened 2024-08-27
Add Yubikey-based challenge for tech port unlocking

mkeeter:
See [RFD 492 § Sketch of an unlock policy](https://rfd.shared.oxide.computer/rfd/492#_sketch_of_an_unlock_policy) for the backstory here.

This adds a new kind of `UnlockChallenge`, which requires that the caller generate an [SSH signature](https://github.com/openssh/openssh-portable/blob/master/PROTOCOL.sshsig) of a particular blob of data, using a trusted key that is baked into the SP's firmware.

(opened as a draft pending testing)


faux-mgs/src/main.rs
```diff
@@ -1471,6 +1585,58 @@ async fn monorail_unlock(
     Ok(())
 }
 
+fn ssh_keygen_sign(
+    pub_key_path: &Path,
+    data: &[u8],
+) -> Result<ssh_key::SshSig> {
+    use std::process::{Command, Stdio};
+    let mut child = Command::new("ssh-keygen")
```
lzrd, line 1593: Use a deterministic path to the executable. See previous comment.

faux-mgs/src/main.rs
```diff
@@ -778,6 +790,31 @@ async fn main() -> Result<()> {
     Ok(())
 }
 
+fn ssh_list_keys() -> Result<Vec<String>> {
+    use std::process::{Command, Stdio};
+    let child = Command::new("ssh-add")
```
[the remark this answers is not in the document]
mkeeter: I switched to talking with the SSH agent directly in e006752c131e, instead of shelling out to the various CLIs

faux-mgs/src/main.rs
mkeeter: Also fixed in e006752c131e !

plotnick (commented):
~~Shelling out to `ssh-*` commands seems ok(ish) for our intended use cases and current users, but not ideal. I think that instead we should use the `agent` module from the [sshauth crate](https://github.com/oxidecomputer/sshauth) (adding any additional features we need, though there shouldn't be many) to communicate directly (over a Unix-domain socket) with the SSH agent. I wouldn't make that blocking, but I think it'd be a good direction to head in.~~

Hah! My internet connection has been flaky all day, and you pushed your commit that does the thing I asked for by the time I posted my comment. Please ignore.

gateway-messages/src/mgs_to_sp.rs
```diff
@@ -322,14 +325,57 @@ pub enum MonorailComponentAction {
...
 )]
 pub enum UnlockResponse {
+    /// Unlocks [UnlockChallenge::Trivial]
     Trivial { timestamp: u64 },
+
+    /// Here's your signature!
+    EcdsaSha2Nistp256 {
+        /// SEC1 encoding of the corresponding public key
+        #[serde(with = "BigArray")]
+        key: [u8; 65],
```
lzrd, line 370: It would be nice to see a symbolic constant, but a quick dive into the elliptic_curve::sec1 crate didn't reveal an answer to me.

lzrd (approved):

faux-mgs/src/main.rs
```diff
@@ -1457,6 +1481,76 @@ async fn monorail_unlock(
         UnlockChallenge::Trivial { timestamp } => {
             UnlockResponse::Trivial { timestamp }
         }
+        UnlockChallenge::EcdsaSha2Nistp256(data) => {
+            // We'll either accept a pubkey path from the command line arguments
+            // (`--key`), or talk to the SSH agent to get a list of keys and
+            // have the user pick one interactively
```
labbott, line 1487: I don't love the interactive feature as it seems much different than what we do elsewhere. Could we do something like what humility does for probes and if there's one probe we use it otherwise we specify by number (humility does `usb-1`, we could do `key-0`)?

mkeeter: Okay, I changed it to either taking a public key path (ending in `.pub`), or doing a substring match on the list of keys (so you could specify `oxide-support-001-auth`)

faux-mgs/src/main.rs
```diff
@@ -1457,6 +1502,79 @@ async fn monorail_unlock(
         UnlockChallenge::Trivial { timestamp } => {
             UnlockResponse::Trivial { timestamp }
         }
+        UnlockChallenge::EcdsaSha2Nistp256(data) => {
+            let keys = ssh_list_keys()?;
+            let pub_key = if keys.len() == 1 {
+                keys[0].clone()
```
jgallagher, line 1510: Kinda nitpicky, but if `pub_key` is `Some(_)` should we check that it matches the one key in this case?

faux-mgs/src/main.rs
```diff
@@ -1457,6 +1502,79 @@ async fn monorail_unlock(
...
+                            if found.is_some() {
+                                bail!("multiple keys contain '{pub_key}'");
+                            }
+                            found = Some(k);
+                        }
+                    }
+                    let Some(found) = found else {
+                        bail!(
+                            "could not match '{pub_key}'; \
+                        use `faux-mgs monorail unlock --list` to print keys"
```
jgallagher, line 1536: nit - rustfmt gave up on indenting this multiline string

jgallagher (approved):
LGTM :+1:

faux-mgs/src/main.rs
mkeeter: Good catch, changed to `if keys.len() == 1 && pub_key.is_none()`

faux-mgs/src/main.rs
mkeeter: fixed

PR #291, opened 2024-09-26
Add `MeasurementKind::Dimensionless`

hawkw:
This commit adds a `CpuTctl` variant to the `MeasurementKind` enum. This
is intended to be used to represent AMD CPU T<sub>ctl</sub> thermal
values, which are a unitless value from 0-100 representing a kind of
"abstract thermal throttling danger level" --- see
oxidecomputer/hubris#1881 for details on why we need to differentiate
this from other temperature measurements.

### DIFF
--- a/gateway-messages/tests/versioning/v16.rs
+++ b/gateway-messages/tests/versioning/v16.rs
@@ -1,0 +1,39 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! This source file is named after the protocol version being tested,
+//! e.g. v01.rs implements tests for protocol version 1.
+//! The tested protocol version is represented by "$VERSION" below.
+//!
+//! The tests in this module check that the serialized form of messages from MGS
+//! protocol version $VERSION have not changed.
+//!
+//! If a test in this module fails, _do not change the test_! This means you
+//! have changed, deleted, or reordered an existing message type or enum
+//! variant, and you should revert that change. This will remain true until we
+//! bump the `version::MIN` to a value higher than $VERSION, at which point these
+//! tests can be removed as we will stop supporting $VERSION.
+
+use super::assert_serialized;
+use gateway_messages::measurement::MeasurementKind;
+use gateway_messages::SerializedSize;
+use gateway_messages::SpResponse;
+
+#[test]
+fn measurement_kinds() {
+    let mut out = [0; SpResponse::MAX_SIZE];
+
+    for (kind, serialized) in [
+        (MeasurementKind::Temperature, &[0]),
+        (MeasurementKind::Power, &[1]),
+        (MeasurementKind::Current, &[2]),
+        (MeasurementKind::Voltage, &[3]),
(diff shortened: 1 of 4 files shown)


jgallagher:
> So, perhaps this ought to be a `CpuTctl` variant or similar, to _specifically_ represent Tctl measurements. If other dimensionless values show up in future, they would then have their own variant. I'm open to being convicned either way --- what do others think?

I think I'd vote for `CpuTctl` Adding `Dimensionless` means future measurement unit/kinds in the future that don't fit into one of the existing variants have somewhere to land without software changes, which is good, but also feels _vague_. Presumably two different sensors with the same `MeasurementKind` are comparable in some way (although that comparison is maybe meaningless and requires more context to know?), but two `Dimensionless` sensors are definitely not comparable without knowing exactly what they are.

But this is an area well outside where I've been any time recently, so I'm happy to defer to others too. @mkeeter might have opinions too; he worked on the SP <-> IPCC <-> host inventory path with @rmustacc.

mkeeter:
I'm also weakly in favor of `CpuTctl`; enum variants are basically free, and it's a weird corner case.

Because this is a change to the SP-MGS protocol, you should also

- Increase [`CURRENT`](https://github.com/oxidecomputer/management-gateway-service/blob/fdf14851a26d918ebac06c3c3c2227be34c99efd/gateway-messages/src/lib.rs#L69) to 16
- Add a new [`tests/versioning/v16.rs`](https://github.com/oxidecomputer/management-gateway-service/tree/fdf14851a26d918ebac06c3c3c2227be34c99efd/gateway-messages/tests/versioning) which tests the serialization of `MeasurementKind::Dimensionless`

This ensures that we get helpful error reporting if there's a mismatch between SP and MGS; specifically, MGS can report "I failed to serialize this message, but it's on v16 while I'm only on v15, so that's not totally surprising".

hawkw:
Okay, I changed the variant to be _just_ for T<sub>ctl</sub>, bumped the version, and added a test.

gateway-messages/tests/versioning/v16.rs
```diff
@@ -0,0 +1,39 @@
...
+//! tests can be removed as we will stop supporting $VERSION.
+
+use super::assert_serialized;
+use gateway_messages::measurement::MeasurementKind;
+use gateway_messages::SerializedSize;
+use gateway_messages::SpResponse;
+
+#[test]
+fn measurement_kinds() {
+    let mut out = [0; SpResponse::MAX_SIZE];
```
mkeeter, line 25: Nit: this could be `MeasurementKind::MAX_SIZE`

mkeeter (approved):
Normally, I open a Hubris PR staged against the MGS branch before merging it (to make sure the Hubris integration doesn't require further changes in MGS).

In this case, the change is simple enough that doesn't seem necessary.

hawkw:
> Normally, I open a Hubris PR staged against the MGS branch before merging it (to make sure the Hubris integration doesn't require further changes in MGS).
> 
> 
> 
> In this case, the change is simple enough that doesn't seem necessary.

Yeah, I would be quite happy to leave this open until the Hubris-side change is done. But, in this case, I also feel like, well...if fixing this in Hubris actually requires more protocol changes, I'll eat a ST-Link live at Oxcon. :)

jgallagher (approved):

PR #297, opened 2024-10-08
Add SP5_HOST_CPU to SpComponent list

mkeeter:
Prep for Grapefruit / Cosmo support in `control-plane-agent`

### DIFF
--- a/gateway-messages/src/lib.rs
+++ b/gateway-messages/src/lib.rs
@@ -369,6 +369,9 @@ impl SpComponent {
     /// The `sp3` host CPU.
     pub const SP3_HOST_CPU: Self = Self { id: *b"sp3-host-cpu\0\0\0\0" };
 
+    /// The `sp5` host CPU.
+    pub const SP5_HOST_CPU: Self = Self { id: *b"sp5-host-cpu\0\0\0\0" };
+
     /// The host CPU boot flash.
     pub const HOST_CPU_BOOT_FLASH: Self = Self { id: *b"host-boot-flash\0" };
 


gateway-messages/src/lib.rs
```diff
@@ -369,6 +369,9 @@ impl SpComponent {
     /// The `sp3` host CPU.
     pub const SP3_HOST_CPU: Self = Self { id: *b"sp3-host-cpu\0\0\0\0" };
 
+    /// The `sp5` host CPU.
+    pub const SP5_HOST_CPU: Self = Self { id: *b"sp5-host-cpu\0\0\0\0" };
```
jgallagher, line 373: Do you think we should move to just `HOST_CPU`? I was expecting someone to care about the difference, but I'm not sure they do.

jgallagher (approved):

gateway-messages/src/lib.rs
mkeeter: I was on the fence here, but am leaning towards not changing the existing name (e.g. it may have ended up in clickhouse for sensor readings?)

jgallagher: Yeah I think we can't really _change_ the name, at least in a single release, but we could add `HOST_CPU` and use that for new things moving forward. (Whether to transition existing uses of `SP3_HOST_CPU` to use `HOST_CPU` instead would be up for debate.)

jgallagher: It seems fine to use this as-is, and if it turns out that these being two different components is annoying, we could change it then. I _think_ the only real user of this is the serial console (where the messages are phrased as "this is the serial console for `SP3_HOST_CPU`).

hawkw: > I was on the fence here, but am leaning towards not changing the existing name (e.g. it may have ended up in clickhouse for sensor readings?)

IIRC, this component doesn't report sensor readings (the CPU T<sub>ctl</sub> temperature measurement is a different component ID). So, I don't _believe_ `SP3_HOST_CPU` is presently in Clickhouse.

PR #306, opened 2024-11-01
Add `faux-mgs current-time` subcommand

### LINKED ISSUE #283: Expose current-time in faux-mgs / MGS
#110 added `MgsRequest::ReadSensor` and `MgsRequest::CurrentTime`. We have a faux-mgs subcommand for `ReadSensor` but are missing `CurrentTime`; we should add that.

jgallagher:
Closes #283

### DIFF
--- a/faux-mgs/src/main.rs
+++ b/faux-mgs/src/main.rs
@@ -235,11 +235,12 @@ enum Command {
         component: SpComponent,
     },
 
-    /// Ask the SP for its current system time (interpreted as human time).
-    CurrentTime,
-
-    /// Ask the SP for its current system time (fetches the raw value).
-    CurrentTimeRaw,
+    /// Ask the SP for its current system time (interpreted as human time or as
+    /// a raw value).
+    CurrentTime {
+        #[clap(short, long, help = "do not interpret returned value")]
+        raw: bool,
+    },
 
     /// Attach to the SP's USART.
     UsartAttach {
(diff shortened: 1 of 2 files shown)


gateway-sp-comms/src/single_sp.rs
```diff
@@ -494,6 +494,11 @@ impl SingleSp {
             .and_then(expect_component_clear_status_ack)
     }
 
+    /// Request that the current system time.
```
mkeeter, line 497:
```suggestion
    /// Request the current system time (in milliseconds since SP boot)
```

gateway-sp-comms/src/single_sp.rs
```diff
@@ -494,6 +494,11 @@ impl SingleSp {
...
+    pub async fn current_time(&self) -> Result<u64> {
+        self.rpc(MgsRequest::CurrentTime).await.and_then(expect_current_time)
```
mkeeter, line 499: Nit: do we want to return a `Duration` here?

jgallagher: We could if we're willing to bake the knowledge here that the SP is always returning ms?

mkeeter: Yeah, I'm on the fence.  On one hand, it's not _required_, but a whole lot of SP firmware assumes it...

jgallagher: Some options:

* Change this to assume millis
* Keep this as-is
* Change this to report both the raw value and interpreted as millis
* Rename this `current-time-raw` and add `current-time` which assumes millis

cbiffle:
FWIW, all Oxide firmware currently assumes that 1 tick == 1 millisecond, and I don't see us changing that. So in any Oxide-specific code this assumption seems reasonable.

gateway-sp-comms/src/single_sp.rs
mkeeter: I like option 4 (having both `current_time_raw` and `current_time`, assuming millis)

jgallagher: Done in c29e480:

```
root@oxz_switch0:/tmp# FAUX_MGS=./john-faux-mgs pilot sp exec -e 'current-time-raw' BRM44220008
Nov 01 19:34:25.321 INFO creating SP handle on interface sidecar1, component: faux-mgs
Nov 01 19:34:25.322 INFO initial discovery complete, addr: [fe80::aa40:25ff:fe05:401%79]:11111, interface: sidecar1, component: faux-mgs
current time (raw): 857253425
root@oxz_switch0:/tmp# FAUX_MGS=./john-faux-mgs pilot sp exec -e 'current-time' BRM44220008
Nov 01 19:34:31.021 INFO creating SP handle on interface sidecar1, component: faux-mgs
Nov 01 19:34:31.022 INFO initial discovery complete, addr: [fe80::aa40:25ff:fe05:401%79]:11111, interface: sidecar1, component: faux-mgs
current time: 9days 22h 7m 39s 126ms
```

faux-mgs/src/main.rs
```diff
@@ -235,9 +235,12 @@ enum Command {
         component: SpComponent,
     },
 
-    /// Ask the SP for its current system time.
+    /// Ask the SP for its current system time (interpreted as human time).
```
mkeeter, line 238: I would find `faux-mgs current-time [--raw]` slightly more ergonomic, but YMMV

mkeeter (approved):

PR #312, opened 2024-11-15
Automatic sizing in `assert_serialized`

mkeeter:
This adds one more allocation (because we can't use the generic parameter, `error: constant expression depends on a generic parameter`), but it makes a bunch of other code less verbose.


hawkw (approved):
Looks good to me --- this is definitely nicer!

PR #316, opened 2024-11-18
Read SP task crash dumps over the network

mkeeter:
This PR adds new messages and functions to read SP task crash dumps over the network and write them to a canonical file, _without_ the corresponding Hubris archive being available at debug time.

These "dry" dumps can then be hydrated with `humility hydrate` (https://github.com/oxidecomputer/humility/pull/520).  In the meantime, they are significantly smaller (RAM only, no ELF or debug info) and easier to pass around (we can using the existing MGS connection instead of talking to `dump-agent` and knowing how `humpty` works).


faux-mgs/src/main.rs
```diff
@@ -455,6 +461,26 @@ enum MonorailCommand {
...
 }
 
+#[derive(Subcommand, Debug, Clone)]
+enum DumpCommand {
+    /// List the number of task crash dumps available
+    Count,
+
+    /// Read a single dump
+    Read {
+        /// File to write the dump
```
jgallagher, line 471: Wrong docstring for this arg?

gateway-messages/src/mgs_to_sp.rs
```diff
@@ -407,3 +409,23 @@ bitflags::bitflags! {
...
+pub enum DumpRequest {
+    /// Returns the number of task crash dumps
+    TaskDumpCount,
+
+    /// Begins reading a task crash dump by index
+    ///
+    /// `key` is used for disambiguation between multiple readers
+    TaskDumpReadStart {
+        index: u32,
+        key: u32,
```
jgallagher, line 425: This is probably fine, but for update uniqueness we use a UUID (converted into a `[u8;16]` at this level). I'm lightly inclined to repeat that here, for the consistency + ease of `Uuid::new_v4()` over importing rand?

gateway-messages/src/mgs_to_sp.rs
```diff
@@ -407,3 +409,23 @@ bitflags::bitflags! {
...
+    },
+
+    TaskDumpReadContinue {
+        key: u32,
```
jgallagher, line 429: Does this need an offset to be idempotent under retries? If we had one, would we need a distinct `ReadStart` command vs sending a `ReadContinue` with offset 0? I guess the later would require resending the task index every time, which isn't ideal...

gateway-sp-comms/src/single_sp.rs
```diff
@@ -1048,6 +1137,168 @@ impl SingleSp {
...
+                },
+            )?;
+            if let Some((header, data)) = r {
+                debug!(
+                    self.log,
+                    "got {} bytes from {:#08x}",
+                    data.len(),
+                    header.address
+                );
+                // The decompressor type must agree with `humpty`
```
jgallagher, line 1264: Should one of the responses (probably to dump start?) have a field for what kind of compression was used, even if that field currently only has one possible value?

jgallagher (commented):
LGTM; some minor questions about the messages.

gateway-sp-comms/src/single_sp.rs
```diff
@@ -1002,6 +1076,21 @@ impl SingleSp {
...
+        key: [u8; 4],
+    ) -> Result<String> {
+        let value = self.read_component_caboose(component, slot, key).await?;
+
+        Ok(if value.is_ascii() {
+            String::from_utf8(value).unwrap()
+        } else {
+            hex::encode(value)
+        })
+    }
```
labbott, line 1092: Do we need a crate version bump for this addition?

faux-mgs/src/main.rs
mkeeter: Yup, fixed in 2bcb9396

gateway-sp-comms/src/single_sp.rs
mkeeter: Good idea, added in 6cf8e7c1

gateway-sp-comms/src/single_sp.rs
mkeeter: I'm not sure why, can you say more?  This is just a helper function to consolidate the "read, then convert to string or hex" pattern from a few places.

labbott: Ah I think I misread. I thought it was a change to `read_component_caboose`  as opposed to a helper function. You are right, we shouldn't need a bump.

mkeeter: (we've also been playing fast and loose with semver in this crate, since we control when we bump the version in Hubris's `Cargo.lock`)

gateway-messages/src/mgs_to_sp.rs
mkeeter: Good catch, I added an explicit sequence number (which also requires support on the SP side)

faux-mgs/src/main.rs
```diff
@@ -419,6 +419,12 @@ enum Command {
         #[clap(subcommand)]
         cmd: MonorailCommand,
     },
+
+    /// List and read per-task crash dumps
+    Dump {
```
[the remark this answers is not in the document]
mkeeter: Probably SP-only?  It's pretty specific to Hubris, and we don't do task dumps on the RoT for security reasons.

jgallagher (approved):

PR #366, opened 2025-03-21
Add Rhai scripting to faux-mgs enabled by feature "rhaiscripting"

lzrd:
I'm just looking for some casual feedback on this. While this PR scratches my own itch for development test automation, it isn't on any critical path.

It should be useful to implement CI against hardware at some point.

This scripting feature is much nicer to use than the bash scripts in the sprot-e2e repo. It was the threat of having to update those that motivated this PR.

See the README.md file in `scripts/README.md`.

This PR has the `rhaiscript` feature on by default.

Thanks to John for some help about a year ago to help me with async recursion.


labbott:
I like the idea of a rhai scripting interface over writing a bunch of shell. I also think there's multiple parts to this: there's the actual rhai interface for faux-mgs and then there's making that interface usable for test scripts. racktest uses rhai but it's slightly different because both the interface and the test scripts are within the same package which is not the case with faux-mgs.

I think the `verify_rot_image` function is a good example of something practical we do want but I don't think `faux-mgs` should be growing that much knowledge of hubris specific functions.  We should figure out what's a better approach for features/functions like that.

Another separate thought: Is this is a sign that we're doing too much with the faux-mgs binary and maybe we need `faux-mgs-lib`? This would let us write something a little bit closer to what racktest does in a separate repository.

lzrd:
I added [Hubtools issue #51](https://github.com/oxidecomputer/hubtools/issues/5). Knowledge about Hubris archives should migrate there.

PR #370, opened 2025-03-27
Ingest ereports from SPs

hawkw:
This pull request implements the MGS side of the SP ereport ingestion
protocol. For more information on the ereport ingestion protocol, refer
to the following RFDs:

- [RFD 520  Control Plane Fault Ingestion and Data Model][RFD 520]
- [RFD 544 Embedded E-Report Formats][RFD 544]
- [RFD 545 Firmware E-Report Aggregation and Evacuation][RFD 545]

In particular, this branch makes the following changes:

- Add types to `gateway-messages` representing the ereport protocol wire
  messages exchanged between MGS and the SP; these are defined in 
  [RFD 545].
- Somewhat substantial refactoring to the `shared_socket` module in
  `gateway-sp-comms`. Currently, the `SharedSocket` code for handling
  received packets is tightly coupled to the control plane agent message
  types. Ereport requests and responses are sent on a separate UDP port.
  Therefore, I've hacked up this code a bit to allow `SharedSocket` to
  be generic over a `RecvHandler` trait that defines how to handle
  received packets and dispatch them to single-SP handlers. This is
  implemented for both the control-plane-agent protocol and, separately,
  for the ereport protocol.
- Actually add said implementation of the ereport protocol, including
  code for decoding ereport packets and a per-SP worker task that tracks
  the metadata sent by the SP and adds it to each batch of ereports.

A corresponding Omicron branch, oxidecomputer/omicron#7903, depends on
this branch and integrates the ereport code into the MGS app binary and
the SP simulator.

[RFD 520]: https://rfd.shared.oxide.computer/rfd/0520
[RFD 544]: https://rfd.shared.oxide.computer/rfd/0544
[RFD 545]: https://rfd.shared.oxide.computer/rfd/0545


gateway-messages/src/ereport.rs
```diff
@@ -0,0 +1,383 @@
...
+/// used to detect restarts.
+///
+/// See [RFD 520 §4.2.2] for details.
+///
+/// [RFD 520 §4.2.2]: https://rfd.shared.oxide.computer/rfd/0520#reporter-crash-recovery
+#[derive(
+    Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, SerializedSize,
+)]
+#[repr(transparent)]
+pub struct RestartId(pub u128);
```
hawkw, line 65: I did consider using the `uuid` crate for this, as it supports `no_std`, but it didn't really seem worth adding another dependency that would have to be compiled in to the Hubris binaries basically just to get UUID-like formatting in `Debug` impls that are only used by MGS...

gateway-messages/src/ereport.rs
hawkw, line 1: I felt that it was nicer to put all the ereport messages (both the SP-to-MGS messages and the MGS-to-SP messages) in their own module, rather than putting some of them in `sp_to_mgs` and others in `mgs_to_sp`. This way, a reader interested in the ereport stuff need only read this module, and a reader interested in the control-plane-agent protocol doesn't have to scroll past ereport messages. Future additions to the ereport protocol would change only the code in this file.

gateway-sp-comms/Cargo.toml
```diff
@@ -27,7 +30,7 @@ thiserror.workspace = true
 tlvc.workspace = true
 tokio.workspace = true
 usdt.workspace = true
-uuid.workspace = true
+uuid = { workspace = true, features = ["v4"] }
```
hawkw, line 33: This was necessary for unrelated reasons: the task dump code added in #316 uses `Uuid::new_v4` in `gateway-sp-comms`, but `gateway-sp-comms` doesn't enable the `"v4"` feature (only `faux-mgs` does). So, this didn't compile for me using the v2 cargo feature resolver.

gateway-sp-comms/src/ereport.rs
```diff
@@ -0,0 +1,752 @@
...
+                        task_names.push(name.clone());
+                        name
+                    }
+                    Some(CborValue::Integer(i)) => task_names
+                        .get(i as usize)
+                        .cloned()
+                        .ok_or(DecodeError::BadTaskNameIndex {
+                            n,
+                            index: i as usize,
+                        })?,
```
hawkw, line 317: @cbiffle this is what I was asking about in https://github.com/oxidecomputer/rfd/pull/849#discussion_r2027376877: right now, this code assumes that when an ereport's task name is an integer, it will always be the index of an ereport that was _earlier in the packet_ than the current one. I wanted to get your confirmation of whether we could rely on that assumption or would need to handle indexes pointing _ahead_ of the current ereport.

gateway-sp-comms/src/shared_socket.rs
```diff
@@ -78,67 +81,89 @@ pub struct BindError {
...
-    single_sp_handlers:
-        Arc<Mutex<FxHashMap<Name, mpsc::Sender<SingleSpMessage>>>>,
+    single_sp_handlers: SpHandlerMap<T>,
     recv_handler_task: JoinHandle<()>,
     log: Logger,
 }
 
-impl Drop for SharedSocket {
+// Hand-rolled `Debug` impl as the message type (`T`) needn't be `Debug` for the
+// `SharedSocket` to be debug.
```
jgallagher, line 93: This seems fine, but I'll plug https://docs.rs/derive-where/latest/derive_where/ since we use it in omicron, if you want to pull it in here too.

gateway-sp-comms/src/single_sp.rs
```diff
@@ -282,15 +284,18 @@ details in `dump.json`."
 pub struct SingleSp {
     interface: String,
     cmds_tx: mpsc::Sender<InnerCommand>,
+    ereport_req_tx: mpsc::Sender<ereport::WorkerRequest>,
```
jgallagher, line 287: I would not do this as a part of this PR, but I'm curious for your thoughts: there are a bunch of places like this one where we're going from one "thing" (in this case, an `mpsc::Sender`) to two "things". The original "one" is always named generically, and the new one is named indicating it's related to ereports. Do you think we should go back and rename the generic things to indicate they're intended for control-plane-agent?

gateway-sp-comms/src/ereport.rs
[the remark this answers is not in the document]
jgallagher: No argument from me. Do you think it would be clearer to move the control-plane-agent stuff to its own submodule too? (Not as part of this PR, but maybe alongside the renamings I suggested in another comment?)

gateway-sp-comms/src/ereport.rs
```diff
@@ -0,0 +1,831 @@
...
+use slog::warn;
+use slog_error_chain::SlogInlineError;
+use std::collections::BTreeMap;
+use std::num::NonZeroU8;
+use thiserror::Error;
+use tokio::sync::mpsc;
+use tokio::sync::oneshot;
+
+pub const SP_PORT: u16 = 0xDEAD;
+pub const MGS_PORT: u16 = 0xDEAF;
```
jgallagher, line 34: Tiny nit - could we write these in decimal? Most OS networking tools deal with ports in decimal, and I've occasionally grepped through codebases to pick out port numbers, and these being in hex would be a bit of a stumbling block.

gateway-sp-comms/src/ereport.rs
```diff
@@ -0,0 +1,831 @@
...
+            // If we have just started up, or we were previously unable to
+            // refresh the SP's metadata, try to do so now.
+            if self.metadata.is_none() {
+                trace!(self.log(), "requesting initial SP metadata...");
+                match self.request_ereports(RestartId(0), Ena(0), 0, None).await
+                {
+                    Ok((restart_id, ereports)) => {
+                        debug_assert!(
+                            ereports.is_empty(),
+                            "we asked for limit 0..."
```
jgallagher, line 100: I'm tempted to give this a pass since it's a `debug_assert!`, but it seems a little fishy: we're asserting something that we have no control over, right (i.e, this assert fails if the remote end is misbehaving)? I'm inclined to say this should be treated as a runtime error instead.

gateway-sp-comms/src/ereport.rs
```diff
@@ -0,0 +1,831 @@
...
+                "attempt" => attempt,
+            );
+            self.socket.send(packet).await?;
+            if let Ok(Some(msg)) = tokio::time::timeout(
+                self.retry_config.per_attempt_timeout,
+                self.socket.recv(),
+            )
+            .await
+            {
+                return decode_packet(restart_id, &mut self.metadata, &msg)
```
jgallagher, line 210: Is it a problem if this is a response to an earlier request? On the CPA side we've seen interleavings like this:

1. MGS sends request A
2. No response after the `per_attempt_timeout`, so MGS resends request A
3. SP sends a response to request A (from step 1)
4. MGS receives the response and returns it to its caller
5. MGS sends request B
6. SP sends a response to request A (from step 2)
7. MGS receives the response and [discards it](https://github.com/oxidecomputer/management-gateway-service/blob/70b551d9b9f9fff0e3c17064e93e59dd2a238f39/gateway-sp-comms/src/single_sp.rs#L2433-L2438) because we know based on the packet header it's not a response to the request we're waiting for

faux-mgs/src/main.rs
```diff
@@ -91,6 +93,11 @@ struct Args {
     #[clap(long)]
     listen_port: Option<u16>,
 
+    /// Ereport port to bind to locally [default: 0 for client commands, 22223
+    /// for server commands]
```
jgallagher, line 97: Does the ereport system have server commands?

jgallagher (commented):
I didn't look particularly closely at the ereport parsing etc details (I'll defer that to you and Cliff, if that's okay). The structural MGS changes look good; just a handful of nits and questions.

gateway-sp-comms/src/ereport.rs
hawkw: The debug assert was really just intended to catch this in testing, but I agree that it is not under MGS' control. We should maybe just return an error to the requester instead, I think you're right about that.

gateway-sp-comms/src/ereport.rs
hawkw: Hmm. So, thinking about this in terms of the ereport protocol, it's not that bad if this happens --- all the messages are idempotent, and if Nexus receives ereports it already received previously, that doesn't cause duplicate events or anything. Every response includes the ENAs and restart ID so we can distinguish between the same ereports twice and new ereports with the same data.

However, it _does_ mean we could reply to a request for ereports with ereports that were intended for a previous request, rather than the ones actually requested by the new request...which "wastes" the new request and results in MGS then dropping the subsequent ereports that were actually intended for that request. So, perhaps we should handle this by adding a sequence number to the request-response pairs the way we do in the control-plane agent protocol. @cbiffle how do you feel about adding a sequence number to these?

gateway-sp-comms/src/ereport.rs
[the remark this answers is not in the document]
hawkw: I'm definitely open to doing that in a subsequent change, I think it seems pretty reasonable (especially if we ever add a third socket/protocol for some other thing). I agree that we shouldn't mess with the control-plane-agent stuff in *this* PR though.

gateway-sp-comms/src/single_sp.rs
hawkw: Yeah, I think it might be worthwhile to do that (especially if we ever add a third port to the management network...). You're right that I just didn't really want to touch all the control-plane-agent code in _this_ PR, but I'd be happy to do it in a separate commit.

gateway-sp-comms/src/ereport.rs
hawkw: Added this in 8a49211667642f2a5f9efdbe7ea19080f056c53c.

PR #390, opened 2025-05-12
Make no-op `SetPowerState` calls a success

hawkw:
Presently, the `SetPowerState` MGS request returns an `IllegalTransition` error when the system is already in the desired power state. This is generally not the desired behavior: the `IllegalTransition` error code represents both power state transitions that will *never* succeed, and cases where the SP was already in the desired power state and we didn't have to do anything (which is a success from the perspective of a maohority of callers). In the cases where this should be treated as an error, it's useful to be able to disambiguate between "you tried to request a transition that is never allowed" and "your request failed because it raced with another request", and in most other cases where callers just want to ensure that the system is in the desired power state, it should be treated as a success. See #270 for details.

This change modifies the gateway-SP protocol to allow the SP to communicate more accurately about no-op power state transitions. `SpHandler`'s `set_power_state` method now returns a `PowerStateTransition` type that indicates whether or not a change occurred, and the `SpResponse` message enum has grown a new `PowerStateUnchanged` variant. We now construct either that variant or the `PowerStateSet` variant (née `SetPowerStateAck`, it's okay to rename variants already in the protocol as it won't change their encoding on the wire) based on whether or not the handler actually indicates that a change occurred. In `gateway-sp-comms`, we then re-construct the `PowerStateTransition` enum fro mthe two possible `SpResposne` messages.

Encoding the no-op case as a new `SpResponse` message was chosen because it's the most backwards-compatible way to do this: in previous versions, the `SpResponse::SetPowerStateAck` message would only be sent if the state transitioned, and an error would be sent otherwise. Adding a new variant for the "no op" case preserves the semantics of the existing message. Alternatively, we could have added a new `SpResponse` variant that holds an additional enum or bool, and deprecate the `SetPowerStateAck` message, but that felt unfortunate, as it left behind a deprecated enum variant that would still need to be handled until it was eventually removed. Also, with two separate variants, we can encode both messages as a single byte instead of two.

On the SP side, oxidecomputer/hubris#2064 lays the groundwork for this by changing the sequencer IPC interface to also indicate no-op power state transitions from `IllegalTransition` errors. An additional change will be necessary to integrate that with the new `gateway-messages` messages in this branch, if the Hubris PR merges first.

### DIFF
--- a/gateway-messages/src/sp_to_mgs.rs
+++ b/gateway-messages/src/sp_to_mgs.rs
@@ -87,7 +87,12 @@ pub enum SpResponse {
     },
     SerialConsoleDetachAck,
     PowerState(PowerState),
-    SetPowerStateAck,
+    /// Indicates that a `SetPowerState` request was performed successfully
+    /// and resulted in a power state transition.
+    ///
+    /// If the SP is already in the desired power state, the
+    /// [`Self::PowerStateUnchanged`] response is returned instead.
+    PowerStateSet,
     ResetPrepareAck,
     // There is intentionally no `ResetTriggerAck` response; the expected
     // "response" to `ResetTrigger` is an SP reset, which won't allow for
@@ -143,6 +148,10 @@ pub enum SpResponse {
     ///
     /// The packet may contain trailing dump data
     Dump(DumpResponse),
+
+    /// Response to a `SetPowerState` request indicating that the system was
+    /// already in the desired power state and no transition occurred.
+    PowerStateUnchanged,
 }
 
 /// Identifier for one of of an SP's KSZ8463 management-network-facing ports.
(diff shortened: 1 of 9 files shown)


gateway-messages/src/sp_to_mgs.rs
```diff
@@ -944,6 +953,24 @@ pub struct UpdateInProgressStatus {
     pub total_size: u32,
 }
 
+/// Represents the result of a successful [`SetPowerState`] request.
+/// [`SetPowerState`]: crate::MgsRequest::SetPowerState
+#[derive(Copy, Clone, PartialEq, Eq)]
+#[repr(u8)]
+pub enum PowerStateTransition {
+    Changed = 1,
```
jgallagher, line 961: Channeling my inner @mkeeter - do we need to skip `0`?

jgallagher (approved):

gateway-messages/src/sp_to_mgs.rs
mkeeter: You probably don't need the `repr(u8)` either 😄

gateway-messages/src/sp_to_mgs.rs
```diff
@@ -87,7 +87,12 @@ pub enum SpResponse {
     },
     SerialConsoleDetachAck,
     PowerState(PowerState),
-    SetPowerStateAck,
+    /// Indicates that a `SetPowerState` request was performed successfully
+    /// and resulted in a power state transition.
+    ///
+    /// If the SP is already in the desired power state, the
+    /// [`Self::PowerStateUnchanged`] response is returned instead.
```
mkeeter, line 94: I don't feel strongly about it, but maybe we should mention that this was formerly known as `SetPowerStateAck`?

hawkw: Good call, I'll add a note!

PR #395, opened 2025-05-30
prototype Wireshark LUA plugin

jgallagher:
As seen at demos; very WIP, but already useful.


wireshark/protofields.lua
```diff
@@ -0,0 +1,379 @@
+-- This file is auto-generated. Any hand edits will be overwritten!
```
jgallagher, line 1: I don't love checking in an autogenerated file, but I think it's correct here, since it makes installing the plugin as simple as "clone repo, add symlink".

wireshark/mgs.lua
```diff
@@ -0,0 +1,59 @@
...
+end
+
+-- main dissector; peel out version and message ID, then
+-- recurse into message-kind-specific dissector
+function mgs.dissector(buffer, pinfo, tree)
+    if buffer:len() < 10 then return end
+
+    pinfo.cols.protocol = mgs.name
+
+    local subtree = tree:add(mgs, buffer(), "MGS Protocol")
```
hawkw, line 37: I might consider calling this "MGS Control Plane Agent Protocol" or something, as it would be nice to add a dissector for the ereport protocol as well. Not a big deal though.

compose-wireshark-plugin/src/main.rs
```diff
@@ -0,0 +1,187 @@
...
+            )?;
+            writeln!(f)?;
+            writeln!(f, "local M = {{}}")?;
+            writeln!(f)?;
+
+            Ok(())
+        }()
+        .with_context(|| {
+            format!("failed to write to {}/protofields.lua", self.output_dir)
+        })
```
hawkw, line 67: turbo nit: this shows up all over the place, maybe it could be stuck in a function or something?

hawkw (approved):
This is super cool! I'll admit I mostly asked to be tagged on the review because I want to eventually add support for the ereport protocol as well, and I wanted a reminder to actually read the code. But, this is really nicely done, and I really had no major concerns except that I think it's probably worth adding some kind of test that fails if the generated Lua files need to be re-generated.

PR #412, opened 2025-07-15
faux-mgs: add `dump-host-flash` subcommand

jgallagher:


faux-mgs/src/main.rs
```diff
@@ -1818,6 +1829,40 @@ async fn run_command(
...
+                    last_log = Instant::now();
+                }
+                let data = sp.read_host_flash(slot, addr).await.with_context(
+                    || format!("failed to read data at offset {addr}"),
+                )?;
+                let to_write = usize::min(data.len(), (size - addr) as usize);
+                fh.write_all(&data[..to_write])
+                    .with_context(|| format!("failed writing to `{output}`"))?;
+                addr += to_write as u32;
+            }
```
labbott, line 1859: consider https://docs.rs/indicatif/latest/indicatif/ for progress? We use this in humility for progress bars

jgallagher:
Last commit added a `--start-address` flag so we can dump everything after sector 0 via

```
dump-host-flash 0 --start-address 65536 33488896 output-file
```

instead of having to dump the whole thing then manually trim or edit sector 0.

PR #434, opened 2025-09-04
faux-mgs: tech port unlock with permslip

plotnick:
In addition to producing signatures over a technician port unlock challenge via `ssh-agent`, also support online signature and response generation via `permslip` (with authn by `ssh-agent`). See oxidecomputer/permission-slip#252.

### DIFF
--- a/faux-mgs/src/main.rs
+++ b/faux-mgs/src/main.rs
@@ -575,6 +575,7 @@ enum MonorailCommand {
             long,
             alias = "online",
             conflicts_with = "list",
+            conflicts_with = "ssh_auth_sock",
             requires = "key"
         )]
         permslip: bool,
--- a/faux-mgs/src/main.rs
+++ b/faux-mgs/src/main.rs
@@ -18,6 +18,7 @@ use futures::StreamExt;
 use gateway_messages::ignition::TransceiverSelect;
 use gateway_messages::ComponentAction;
 use gateway_messages::ComponentActionResponse;
+use gateway_messages::EcdsaSha2Nistp256Challenge;
 use gateway_messages::IgnitionCommand;
 use gateway_messages::LedComponentAction;
 use gateway_messages::MonorailComponentAction;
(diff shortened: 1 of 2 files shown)


faux-mgs/src/main.rs
```diff
@@ -2015,6 +1962,123 @@ async fn monorail_unlock(
...
+        .arg(key_name)
+        .arg("--sshauth")
+        .arg("--kind=tech-port-unlock-challenge")
+        .stdin(Stdio::piped())
+        .stdout(Stdio::piped())
+        .stderr(Stdio::inherit())
+        .spawn()
+        .context(
+            "unable to execute `permslip`, is it in your PATH and executable?",
+        )?;
```
labbott, line 1983: I'll defer to others but I have a light preference to also allow overloading the path via environment variable e.g. `PERMSLIP` if I need to test something

gateway-messages/src/sp_impl.rs
```diff
@@ -417,6 +420,15 @@ pub trait SpHandler {
...
     fn get_host_flash_hash(&mut self, slot: u16) -> Result<[u8; 32], SpError>;
+
+    /// Unlocks the tech port if the challenge and response are compatible
+    fn unlock(
+        &mut self,
+        vid: Self::VLanId,
+        challenge: UnlockChallenge,
+        response: UnlockResponse,
+        time_sec: u32,
+    ) -> Result<(), MonorailError>;
```
labbott, line 431: What's the advantage of moving this into the trait? Is this to avoid the extra monorail step?

faux-mgs/src/main.rs
plotnick: Good idea, added in 68df5ec.

gateway-messages/src/sp_impl.rs
plotnick: I have not worked in this code base before, so it may just be a mistake. But I was guided by the module-level doc-comment `//! Behavior implemented by both real and simulated SPs.` I figured that with oxidecomputer/omicron#8994, the trait would now be the right place for this operation.

labbott (approved):
LGTM

faux-mgs/src/main.rs
```diff
@@ -555,17 +556,28 @@ enum MonorailCommand {
...
+
+        /// Use the Online Signing Service with `permslip`
+        #[clap(
+            short,
+            long,
+            alias = "online",
+            conflicts_with = "list",
+            requires = "key"
+        )]
+        permslip: bool,
```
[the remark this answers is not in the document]
plotnick: Sure thing, added in b79fe36.

Issue #450, opened 2025-09-25
Consider returning an error if index exceeds component count


mkeeter:
In both `MgsRequest::ComponentDetails` and `MgsRequest::Inventory`, we clamp the index if it exceeds the maximum number of component details or device count (respectively).  This makes me suspicious, and I think we should return `SpError::RequestUnsupportedForComponent` instead.

@jgallagher @hawkw any thoughts?

PR #457, opened 2025-10-06
Return errors on out-of-range indexing

mkeeter:
Fixes https://github.com/oxidecomputer/management-gateway-service/issues/450

I'm pretty confident that no one is sending out-of-range indices, because the previous `min` expressions all had an off-by-one error 🙃 

I'll test this out on actual hardware tomorrow morning.


jgallagher (approved):
Thanks :facepalm:

PR #467, opened 2025-10-28
Improve ignition target selection in `faux-mgs`

mkeeter:
Ignition commands in `faux-mgs` require specifying an "ignition target", which has to be looked up in [RFD 144 § Switch port map](https://rfd.shared.oxide.computer/rfd/144#_switch_port_map).  This isn't the same as cubby number, which is annoying.

This PR makes it easier to deal with:
```
Get the ignition state for a port (only valid if the SP is an ignition controller)

Usage: faux-mgs --interface <INTERFACE> ignition [OPTIONS]

Options:
  -t, --target <TARGET>    Ignition target
  -c, --cubby <CUBBY>      Sled cubby (0-31)
  -s, --sidecar <SIDECAR>  Sidecar ('local' or 'peer')
  -p, --psc <PSC>          PSC index (0-1)
  -a, --all                All targets
  -h, --help               Print help
```

This set of arguments is common to `ignition`, `ignition-link-events`, `clear-ignition-link-events`; `ignition-command` supports all the flags except `--all`.

Opening as a draft because `racktest` uses `faux-mgs ignition all` [here](https://github.com/oxidecomputer/pilot/blob/main/racktest/jobs/ignition.rhai#L111), and we should figure out how to coordinate with its release.  We _could_ keep `all` as a special-cased target (instead of `--all`), but it's kinda annoying...

### DIFF
--- a/faux-mgs/src/main.rs
+++ b/faux-mgs/src/main.rs
@@ -176,12 +176,14 @@ enum Command {
     /// controller).
     Ignition {
         #[clap(flatten)]
-        sel: IgnitionBulkSelector,
+        sel: IgnitionBulkSelectorWithCompatibilityShim,
     },
 
     /// Send an ignition command for a single target port (only valid if the SP
     /// is an ignition controller).
     IgnitionCommand {
+        // We can't use the compatibility shim here because it has an optional
+        // positional argument, which can't be followed by `command`
         #[clap(flatten)]
         sel: IgnitionSingleSelector,
         #[clap(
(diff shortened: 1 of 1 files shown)


faux-mgs/src/main.rs
```diff
@@ -643,6 +635,138 @@ impl std::fmt::Display for CfpaSlot {
...
+            return Ok(None);
+        };
+        Ok(Some(t))
+    }
+}
+
+/// `IgnitionSelector` that also supports `--all`
+#[derive(Clone, Debug, clap::Args)]
+pub struct IgnitionBulkSelector {
+    #[clap(flatten)]
```
jgallagher: Does this preclude having a `conficts_with_all = ["all"]`? Could we get that effect by making `IgnitionBulkSelector` an enum, or would that force us to have a subcommand to specify the variant?

mkeeter:
After chatting with @jclulow, I restored the support for `faux-mgs ignition all` and `faux-mgs ignition [integer]`.  The former is used in `racktest`; I can't find any evidence for the latter being used in scripts.

Due to clap's limitations on optional arguments, I can't support both `faux-mgs ignition-command --cubby=0 power-on` and `faux-mgs ignition-command 0 power-on`, so I'm going to break backwards compatibility here.  Searching through the entire `oxidecomputer` organization, there are no uses of `ignition-command`, `ignition-link-events`, or `clear-ignition-link-events`.

jgallagher (approved):

PR #471, opened 2025-11-19
add ability to set ignition always_transmit

Aaron-Hartwig:


mkeeter:
Two notes:

- Because this is a protocol change, we should bump `version::CURRENT` to the next integer and add serialization tests for the new message type in `gateway-messages/tests/versioning/vXX.rs`.
- In modern times, we vaguely try to avoid adding more messages to the top-level `Message`.  For something like this, I'd be tempted to add a new `IGNITION` component type, then put this into a new `ComponentAction::Ignition` variant.

(paging @jgallagher for a second opinion on the latter)

jgallagher:
> * Because this is a protocol change, we should bump `version::CURRENT` to the next integer and add serialization tests for the new message type in `gateway-messages/tests/versioning/vXX.rs`.

+1. See https://github.com/oxidecomputer/management-gateway-service/pull/445 for a recent example.
 
> * In modern times, we vaguely try to avoid adding more messages to the top-level `Message`.  For something like this, I'd be tempted to add a new `IGNITION` component type, then put this into a new `ComponentAction::Ignition` variant.

Hmm. I'm inclined to suggest adding a new variant to [`IgnitionCommand`](https://github.com/oxidecomputer/management-gateway-service/blob/f227b598d61b86f802ead2092076576059f36373/gateway-messages/src/mgs_to_sp.rs#L288-L292) instead, something like `IgnitionCommand::AlwaysTransmit { enabled: bool }`. I think that fits in with the other bits we need:

* the `target` comes along for free in the existing `MgsRequest::IgnitionCommand { target, command }`
* we'd get back the existing `IgnitionCommandAck` with no info (just like `IgnitionAlwaysTransmitAck` has no info)

Aaron-Hartwig:
@jgallagher much cleaner! Thanks for the feedback.

PR #385, opened 2025-04-30
Add --transient flag to component-active-slot

lzrd:
When setting a RoT Hubris bank as persistent or transient either the -t or -p flag must be used in conjunction with the -s flag.

```
Get or set the active slot of a component (e.g., `host-boot-flash`).

Except for component "stage0", setting the active slot can be viewed as an atomic operation.

Setting "stage0" slot 1 as the active slot initiates a copy from slot 1 to slot 0 if the contents of
slot 1 still match those seen at last RoT reset and the contents are properly signed.

Power failures during the copy can disable the RoT. Only one stage0 update should be in process in a
rack at any time.

Usage: faux-mgs --interface <INTERFACE> component-active-slot [OPTIONS] <COMPONENT>

Arguments:
  <COMPONENT>

Options:
  -s, --set <SLOT>
          set the active slot

  -p, --persist
          persist the active slot to non-volatile memory

  -t, --transient
          Only valid with component "rot": Prefer the specified slot on the next soft reset

  -h, --help
          Print help (see a summary with '-h')
```

Test runs:
```
Run #1: cargo -q run -- -l crit --interface enp5s0 component-active-slot            rot
1
Run #2: cargo -q run -- -l crit --interface enp5s0 component-active-slot         -t rot
error: the following required arguments were not provided:
  --set <SLOT>

Usage: faux-mgs --interface <INTERFACE> component-active-slot --set <SLOT> <--persist|--transient> <COMPONENT>

For more information, try '--help'.
Run #3: cargo -q run -- -l crit --interface enp5s0 component-active-slot      -p    rot
error: the following required arguments were not provided:
  --set <SLOT>

Usage: faux-mgs --interface <INTERFACE> component-active-slot --set <SLOT> <--persist|--transient> <COMPONENT>

For more information, try '--help'.
Run #4: cargo -q run -- -l crit --interface enp5s0 component-active-slot      -p -p rot
error: the argument '--persist' cannot be used multiple times

Usage: faux-mgs --interface <INTERFACE> component-active-slot [OPTIONS] <COMPONENT>

For more information, try '--help'.
Run #5: cargo -q run -- -l crit --interface enp5s0 component-active-slot      -p -t rot
error: the argument '--persist' cannot be used with '--transient'

Usage: faux-mgs --interface <INTERFACE> component-active-slot --set <SLOT> --persist <COMPONENT>

For more information, try '--help'.
Run #6: cargo -q run -- -l crit --interface enp5s0 component-active-slot      -t -t rot
[...]

PR #475, opened 2025-12-17
Fix component-active-slot to not require --persist or --transient

### LINKED ISSUE #474: #385 breaks non-persistent host flash update
While working in the field, we had received a newer faux-mgs and it turns out it is now impossible to do a non-persistent update to the host flash. From a glance this appears to be due to #385. There are two problems here:
 
1. Something has made it so the `component-active-slot` requires one to specify either `--persistent` or `--transient`. This breaks pilot usage.
2. We do non-persistent host SP ROM updates all the time. The check added by #385 claims this can only work on the RoT. This isn't true, but is not helpfully enforced.

(1) is an unforced flag day. The command should have ideally just assumed it was transient because it always had been transient and not required us to specify it. Unfortunately it doesn't matter because of (2). This check needs to be changed and realistically enforcement should not happen at faux-mgs and should instead have happened in the relevant components that do this activate.

This is an example of it failing after adjusting for the requirement that one must specify `--transient`.

```
rm@cadmium ~ $ /usr/bin/faux-mgs --interface dut0 --discovery-addr [fe80::aa40:25ff:fe04:1285]:11111 component-active-slot host-boot-flash -s 1 --transient
Dec 17 00:16:29.497 INFO creating SP handle on interface dut0, component: faux-mgs
Error: The --transient (-t) flag is only allowed for the 'rot' component, not for host-boot-flash
```

lzrd:
PR #385 introduced a regression that required either --persist or --transient when setting the active slot. This broke pilot, which relies on the default behavior where no flag means transient.

Additionally, the --transient flag was incorrectly restricted to only the 'rot' component, but host-boot-flash also supports non-persistent slot changes.

Changes:
- Remove the switch_duration argument group requirement from --set
- Remove the rot-only restriction on --transient
- Add tests for argument parsing

Fixes #474

### DIFF
--- a/faux-mgs/src/main.rs
+++ b/faux-mgs/src/main.rs
@@ -1492,7 +1492,12 @@ async fn run_command(
                 ]))
             }
         }
-        Command::ComponentActiveSlot { component, set, persist, .. } => {
+        Command::ComponentActiveSlot {
+            component,
+            set,
+            persist,
+            transient: _,
+        } => {
             if let Some(slot) = set {
                 sp.set_component_active_slot(component, slot, persist).await?;
                 if json {


lzrd:
Output from [update-test.sh](https://github.com/user-attachments/files/24204674/update-test.sh)

```
$ ./update-test.sh  -i wlp192s0
Building faux-mgs...

=== Pilot use cases (must pass) ===
  get slot: host-boot-flash                                    PASS
  set slot transient (default): -s 0                           PASS
  set slot transient (default): --set 0                        PASS
  set slot persistent: -s 0 -p                                 PASS
  set slot persistent: --set 0 --persist                       PASS
  rot persistent (operate.rs): --set 0 --persist               PASS

=== Explicit --transient flag ===
  host-boot-flash: -s 0 --transient                            PASS
  host-boot-flash: -s 0 -t                                     PASS
  rot: -s 0 --transient (SP rejects)                           PASS
  sp: -s 0 --transient (SP rejects)                            PASS

=== Invalid argument combinations (must fail) ===
  --transient without --set                                    PASS
  --persist without --set                                      PASS
  --persist and --transient together                           PASS
  duplicate --persist                                          PASS
  duplicate --transient                                        PASS

=== Summary: 15 passed, 0 failed ===
```

lzrd:
I had not included `pilot` use cases when I made the CLI change in May. Fixed.

PR #488, opened 2026-04-27
gateway-messages: add `HAS_VPD` and `IS_PMBUS` capabilities

hawkw:
This commit adds new `HAS_VPD` and `IS_PMBUS` bits to the `DeviceCapabilities` bitfield, which will indicate when a device has vital product data, and when a device is a PMBus device, respectively. This is being added separately from actually implementing any of these capabilities mostly in order to avoid future merge conflicts here.

### DIFF
--- a/gateway-messages/src/sp_to_mgs.rs
+++ b/gateway-messages/src/sp_to_mgs.rs
@@ -906,6 +906,15 @@ bitflags! {
         const HAS_MEASUREMENT_CHANNELS = 1 << 1;
         const HAS_SERIAL_CONSOLE = 1 << 2;
         const IS_LED = 1 << 3;
+        /// The device has vital product data, such as a part numbe and serial
+        /// number
+        const HAS_VPD = 1 << 4;
+        /// This is a PMBus device.
+        ///
+        /// This implies that the device `HAS_MEASUREMENT_CHANNELS` and
+        /// `HAS_VPD`, but also indicates that the device has PMBus status
+        /// registers.
+        const IS_PMBUS = 1 << 5;
         // MGS has a placeholder API for powering off an individual component;
         // do we want to keep that? If so, add a bit for "can be powered on and
         // off".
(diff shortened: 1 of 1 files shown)


gateway-messages/src/sp_to_mgs.rs
```diff
@@ -906,6 +906,15 @@ bitflags! {
         const HAS_MEASUREMENT_CHANNELS = 1 << 1;
         const HAS_SERIAL_CONSOLE = 1 << 2;
         const IS_LED = 1 << 3;
+        /// The device has vital product data, such as a part numbe and serial
```
mkeeter:
```suggestion
        /// The device has vital product data, such as a part number and serial
```

mkeeter (approved):

gateway-messages/src/sp_to_mgs.rs
```diff
@@ -906,6 +906,15 @@ bitflags! {
...
         const IS_LED = 1 << 3;
+        /// The device has vital product data, such as a part number and serial
+        /// number
+        const HAS_VPD = 1 << 4;
+        /// This is a PMBus device.
+        ///
+        /// This implies that the device `HAS_MEASUREMENT_CHANNELS` and
+        /// `HAS_VPD`, but also indicates that the device has PMBus status
+        /// registers.
+        const IS_PMBUS = 1 << 5;
```
evan-oxide, line 917: Does this mean we're promising that if the `IS_PMBUS` bit is set, then the `HAS_MEASUREMENT_CHANNELS` and `HAS_VPD` bits will also be set? Or that if the client sees that `IS_PMBUS` is set it should ignore those other two bits and assume the device has those capabilities?

hawkw: I expect that the client will always use `HAS_MEASUREMEMT_CHANNELS` and `HAS_VPD` to detect those capabilities, rather than the PMBus bit. The client side code will generally be simpler this way since it doesn't have to check for the presence of either `IS_PMBUS` *or* `HAS_MEASUREMENT_CHANNELS` when trying to determine if a thing is a sensor. This was more a note for human readers that PMBus devices will always have those additional capabilities. I think these docs could make this clearer by explicitly noting you should still use the other bits to detect those capabilities, I'll clarify that.

PR #489, opened 2026-04-28
gatewasy-messages: clarify `IS_PMBUS` capability detection

hawkw:
@evan-oxide had some [questions][1] about the comment I added in #488, so I thought we could probably be more explicit about how the `IS_PMBUS` bit should be used. Hopefully, this makes it clearer?

[1]:
https://github.com/oxidecomputer/management-gateway-service/pull/488#discussion_r3149808924


evan-oxide (approved):
This is clearer, thanks!

PR #492, opened 2026-05-14
Allow SP to report reason for power state change

evan-oxide:
This addresses https://github.com/oxidecomputer/hubris/issues/2335 by allowing the SP to report the reason for a power state change. It adds `MgsRequest::GetPowerStateWithReason` that contains the reason for and timestamp of the last change.


faux-mgs/src/main.rs
```diff
@@ -2023,15 +2024,24 @@ async fn run_command(
...
-                let state = sp
-                    .power_state()
+                let PowerStateWithReason { state, reason, since } = sp
+                    .power_state_with_reason()
                     .await
                     .context("failed to get power state")?;
-                info!(log, "SP power state = {state:?}");
+                info!(
+                    log,
+                    "SP power state = {state:?}, reason = {reason:?}, since = {since:?}"
```
hawkw, line 2033: take it or leave it, but perhaps we ought to explicitly indicate that `since` is a Hubris uptime timestamp in milliseconds here? otherwise it will just be printed as a number.

even
```suggestion
                    "SP power state = {state:?}, reason = {reason:?}, since = {since:?}ms"
```
is probably fine

hawkw (approved):

PR #495, opened 2026-05-19
Add MGS API for getting Host Panic/Host Boot Failure messages

jamesmunns:
This is a component of https://github.com/oxidecomputer/hubris/pull/2518 ~and is probably not totally ready for feedback yet~.

This is a part of implementing https://github.com/oxidecomputer/hubris/issues/2504.


gateway-messages/src/mgs_to_sp.rs
```diff
@@ -243,8 +243,41 @@ pub enum MgsRequest {
     /// different than the current active slot (see `ComponentGetActiveSlot`).
     ComponentGetPersistentSlot(SpComponent),
 
-    /// Request the STATUS registers of a PMBus device, indexed by power rail name.
+    /// Request the STATUS registers of a PMBus device, indexed by power rail
+    /// name.
     GetPmbusStatus(PowerRailName),
+
+    /// Request for Host Panic Payload
+    GetHostPanicPayload {
```
jgallagher, line 251: With apologies for how tedious this is: since we're adding to the protocol, can we bump the version and add tests that ensure we don't break compatibility in future changes? See #482 for a simple one, or #392 for a somewhat more involved one.

Also it looks like #497 should have had this also 😬. We could either PR a version bump + tests for that separately, then come back to this PR, or roll both changes into one version bump?

faux-mgs/src/main.rs
```diff
@@ -2524,6 +2529,141 @@ async fn run_command(
...
+                    .get_host_panic_payload(
+                        Some(HostInfoRequest {
+                            offset: total.len() as u32,
+                            seqno,
+                        }),
+                        512,
+                    )
+                    .await?;
+
+                // TODO: If either of these change, it would mean that the host
```
jgallagher: I think the TODOs and having to choose a buffer length are signs that `sp.get_host_panic_payload()` is too low-level for what we'd like from the `gateway-sp-comms` interface. It's the common interface between faux-mgs and real MGS, and ideally those behave more-or-less identically. If we've got logic here that's specifically documented as "the control plane should do this differently", that's not true anymore.

Many of the `gateway-sp-comms` methods are simple 1-to-1 wrappers around MGS <-> SP messages, but there are several that wrap multiple messages to present a nicer interface to the control plane; e.g., [`get_component_details()`](https://github.com/oxidecomputer/management-gateway-service/blob/da0ac17b9d465db5b1f7a2597d068ae091a89b22/gateway-sp-comms/src/single_sp.rs#L579-L584) sends multiple requests necessary to collect all the information (similar to this interface I think), or [`start_update()`](https://github.com/oxidecomputer/management-gateway-service/blob/da0ac17b9d465db5b1f7a2597d068ae091a89b22/gateway-sp-comms/src/single_sp.rs#L760-L763), which drives an entire update.

From reading the faux-mgs code here, I think the interface I'd propose is `sp.get_host_panic_payload() -> Result<Vec<u8>, _>` (no arguments, return the entire panic message), which handles picking a reasonable buffer size and embeds this while loop. If we go this route we don't want these assertions though - probably reasonable to return a specific error for "host panic'd while we were reading the panic message" and then let the caller retry?

jamesmunns: I think that's reasonable! I can push down the "get all" logic one layer so that `gateway-sp-comms` handles more of this autonomously.

Just checking for the return type:

* `Result<Option<Vec<u8>>, _>` is cool, you don't want the "sequence number" to be surfaced up? (it is a rough indicator of "how many times has the system panicked/bootfailed since the last SP reboot")
* You want `Vec<u8>`, instead of `String`? The data *should* always be UTF-8, but we can let the caller decide what to do if we got a payload but it was corrupted (or the host gave us bad UTF-8 data).

jamesmunns: Oh, `bootfail` also has a `u8` "failure reason", I'm guessing we *do* want to pass that up as well? (panic doesn't have this).

jgallagher: Ah, yes, sorry, definitely include more than just the message itself if there's more relevant information. I'd include both the sequence number and the failure reason (and any other bits if I missed those).

> You want Vec<u8>, instead of String? The data should always be UTF-8, but we can let the caller decide what to do if we got a payload but it was corrupted (or the host gave us bad UTF-8 data).

I had the same internal discussion before posting that. 😅 I could see reasonable arguments for:

* `Vec<u8>` (fewer error cases to handle in `gateway-sp-comms`, always reports the exact data from the host)
* `String` via `from_utf8` (fails if we get corrupt / bad data, which seems not great)
* `String` via `from_utf8_lossy` (doesn't fail, but is lossy)

I don't think I'd pick "`String` via `from_utf8`". If we really expect these to always be UTF8, I'd be fine with "`String` via `from_utf8_lossy`", since then at least we'll always get as much of a printable thing as there is.

faux-mgs/src/main.rs
```diff
@@ -2524,6 +2529,141 @@ async fn run_command(
...
+                // is to just panic, if we wanted to be really fancy we could
+                // put this whole match arm in an outer loop and gracefully
+                // retry. If you are taking this impl for real control plane
+                // things, consider doing that, maybe with some upper bound of
+                // retries!
+                //
+                // The SP can only store one host panic at a time, so there's
+                // no way to retrieve an older panic after it has been
+                // overwritten.
+                assert_eq!(seqno, res.seqno);
```
jgallagher: This comment is not super relevant given my suggestion above will get rid of this, but just in general: I don't think `assert!` is ever correct in this context - it should be reserved for programmer error / invariant violation, and I think we'd treat any report of a faux-mgs panic as a bug in faux-mgs. 

This is much more of an operational error akin to an I/O error reading a file or somesuch. As noted above I think we don't need to necessarily go all the way to a retry-loop-with-a-cap; it's fine to return an error (again akin to bailing out on an I/O error partway through reading a file).

jamesmunns: Gotcha! It's definitely not something I'd do in a library, but it seemed exceptional enough to be acceptable for a binary tool. This will go away once it gets consumed by the lib API, but I will keep that in mind for the future!

jgallagher: Yeah, gotcha. I think some missing context here is that while this is a binary tool, in practice it has been wrapped up by _other_ binary tools and is therefore used as a quasi library. This is kind of an unfortunate sequence, but does make the "probably shouldn't panic" more reasonable - I think faux-mgs is executed by other tools much, much more often than it's executed directly by a human.

faux-mgs/src/main.rs
jamesmunns: Okay, I am back to report that I am a liar! Panic and Bootfail data are in fact *not* strings, but instead a specific format, for which a parser is provided in [`ipcc-rs/ipcc-data`](https://github.com/oxidecomputer/ipcc-rs/tree/main/ipcc-data), and was defined by RFD0316.

We don't already seem to depend on `ipcc-rs` here yet. I think there are probably two reasonable-ish options:

* Just return the `Vec<u8>`, allow the control plane to depend on both `gateway-sp-comms` and `ipcc-data`.
* Add a dep on `ipcc-data`, and return an `ipcc_data::HSSPanic` instead of a `Vec<u8>`.

I'd *probably* lean towards the former, just because it makes dep versioning less messy to defer integration, but it does still leave this API as "some assembly required".

jgallagher: Yeah, `Vec<u8>` seems fine - MGS can be "just a transport" in this case, and it's up to the control plane and host to coordinate on how to interpret it.

That said, we probably want to add `ipcc-data` as a dep to `faux-mgs`, so it can do similar interpretation? Otherwise it's stuck with just a blob?

jamesmunns: Yup, it seems much more reasonable to add to `faux-mgs` than `gateway-sp-comms`, imo.

jamesmunns: Added

faux-mgs/src/main.rs
jamesmunns: Removed

faux-mgs/src/main.rs
```diff
@@ -2524,6 +2530,68 @@ async fn run_command(
...
+            out.push("  contents:".to_string());
+            out.push(String::new());
+            out.extend(format!("{panic_msg:#?}").lines().map(str::to_string));
+            Ok(Output::Lines(out))
+        }
+        Command::GetBootFail => {
+            let bf_payload = sp.get_host_bootfail_payload().await?;
+            // Reason meaning, according to
+            // https://rfd.shared.oxide.computer/rfd/0316
+            let reason = match bf_payload.reason {
```
jgallagher, line 2557: Should `ipcc_data` have an enum that handles this mapping for us a la `PanicData`?

(If so, certainly doesn't need to block this PR, but we could file an issue there?)

faux-mgs/src/main.rs
```diff
@@ -2524,6 +2530,68 @@ async fn run_command(
...
+            out.push("Got bootfail Data:".to_string());
+            out.push(format!("  total bytes:     {}", bf_payload.total_len));
+            out.push(format!("  sequence number: {}", bf_payload.seqno));
+            out.push(format!("  reason:          {reason}"));
+            out.push(format!("  boot slot:       {:?}", bf_payload.slot));
+            out.push("  contents:".to_string());
+            out.push(String::new());
+
+            // Hand-rolled hexdump, as there is currently no IPCC HSSBootFail
+            // parser in ipcc-data
```
jgallagher, line 2580: Ditto on filing issues against ipcc-rs for things we wish it had

gateway-sp-comms/src/single_sp.rs
```diff
@@ -1458,6 +1459,173 @@ impl SingleSp {
...
     }
+
+    pub async fn get_host_panic_payload(&self) -> Result<HostPanicPayloadData> {
+        // Get the first segment, if any
+        //
+        // TODO: minmax buffer sizing? We have a normal sized UDP frame (15xx
+        // bytes?), with some overhead, 1k might be reasonable here, but
+        // there's probably not that much speed benefit to halving the
+        // number of frames sent as this isn't done in a "hot" loop.
+        let res = self.get_host_panic_payload_raw(None, 512).await?;
```
jgallagher: Tiny style nit: can we put 512 in a (private to this method) `const` and attach this comment to that const, so we don't duplicate it here and within the loop below?

gateway-sp-comms/src/single_sp.rs
```diff
@@ -1458,6 +1459,173 @@ impl SingleSp {
...
+            //
+            // Usually, the sequence number should change, but if for some
+            // reason the SP panicked as well (and reset its sequence number),
+            // and then the host panicked again, the total length could
+            // theoretically change as well.
+            let changed = (seqno != res.seqno)
+                || (ttl_bytes != res.total_len)
+                || (slot != res.slot);
+            if changed {
+                return Err(CommunicationError::HostDataSequenceChanged);
```
jgallagher: It's probably worth including the specific value(s) that changed in this error so we get as much info from an unexpected / unusual error as we can. I don't have a strong opinion between stuffing all 6 values (each before/after) into this error variant, or splitting the error variant into three with 2 values each; seems like each option has some pros/cons.

gateway-sp-comms/src/single_sp.rs
```diff
@@ -1458,6 +1459,173 @@ impl SingleSp {
...
+            // RIGHT as we were asking about it.
+            //
+            // The SP can only store one host panic at a time, so there's
+            // no way to retrieve an older panic after it has been
+            // overwritten. The caller should retry until this succeeds.
+            //
+            // Usually, the sequence number should change, but if for some
+            // reason the SP panicked as well (and reset its sequence number),
+            // and then the host panicked again, the total length could
+            // theoretically change as well.
```
jgallagher: I guess this means if the SP and host both panicked together, we could think we got back a successful panic message but it was actually split between two? Something like:

1. SP is on its initial seq number (0? 1? whatever)
2. Host panics with an arbitrary slot / total_len
3. MGS gets partway through this loop
4. SP and host both panic - SP panic means we get the same seq number as in 1; suppose the host panic happens to produce the same slot / total_len as 2
5. We continue reading and don't realize anything is amiss

Does this seem right? (Even if so, step 4 seems exceedingly unlikely in practice.)

gateway-sp-comms/src/single_sp.rs
```diff
@@ -1458,6 +1459,173 @@ impl SingleSp {
...
+        total.truncate(ttl_bytes);
+
+        // Request the entire contents, one chunk at a time
+        while total.len() < ttl_bytes {
+            // Get the NEXT chunk of data, after the part(s) that we already
+            // have received.
+            let res = self
+                .get_host_bootfail_payload_raw(
+                    Some(HostInfoRequest { offset: total.len() as u32, seqno }),
+                    512,
```
jgallagher: (Same comments as `get_host_panic_payload()` w.r.t. making this 512 a const and adding more detail to the error variant)

jgallagher (commented):
Thanks, the changes look great. Mostly just some small nits/questions.

gateway-messages/src/mgs_to_sp.rs
jamesmunns: Resolved by #498

faux-mgs/src/main.rs
jamesmunns: Yep, I'll open an issue for that, and the other comment below for the payload.

gateway-sp-comms/src/single_sp.rs
jamesmunns: I can do that, however I might counter that regardless of what has changed, the actual before/after isn't *meaningful*. Any change at this point means "there has been a change, and what we were trying to get before is not available anymore, and you should now treat what you were trying to get before as "gone", and start trying to get the new value.

I'd be happy to add the 6 fields to this error variant, but IMO there's nothing to *do* with that data. That being said, it could be useful for the sake of logging, in case there is some kind of regression in the logic on either side in the future!

Happy to take notes on taste here, though that's why I didn't include it previously.

jgallagher: > I'd be happy to add the 6 fields to this error variant, but IMO there's nothing to do with that data. That being said, it could be useful for the sake of logging, in case there is some kind of regression in the logic on either side in the future!

Yeah, this is all I was imagining too. I think we've been burned a handful of times by "error we never expected to see in practice is now being logged, and we wish it had more details". This being just for logging is largely why I don't have a strong opinion on 6 fields (or 8? if we need to include `reason` for get_host_bootfail_payload()`? although at that point maybe splitting it into two variants makes sense) vs multiple error variants.

gateway-sp-comms/src/single_sp.rs
jamesmunns: > suppose the host panic happens to produce the same slot / total_len as 2

Yes, that is correct, given that 4 happens.

AFAIK, the only way to strengthen this would be to *also* include the SP's "boot ID" which is a multi-byte random number we generate on every boot, in every bootfail/host panic message. I think this would then make 4's criteria "and our hardware RNG generates the same number between the two boots", which is likely even closer to "impossible". Or we have some kind of stateful "has the 0th chunk been obtained before asking for the 1..th chunk".

I'm unsure if that is a higher robustness barrier than is included in other paginated responses, I'd have to go look if other ones include that!

jgallagher: I don't think any other responses include the boot ID (or if there are, it's not very many). I suspect most paginated responses fall into one of three categories:

1. It'll fail if the SP resets partway through (I don't know if any paginated _responses_ fall into this category, but there's a set of paginated _requests_ that do: update delivery from MGS -> SP will fail if the SP resets partway through due to in-memory state on the SP)
2. It'll succeed and be fine (e.g., if we're reading the host flash and the SP resets partway through, we can still continue reading the host flash from wherever we were)
3. It'll succeed and be fine as long as the reset wasn't due to an update that changes that specific paginated response (e.g., if we're reading the inventory and the SP resets partway through, we can continue from where we were as long as the reset didn't boot us into an image that changed the order / number of inventory items; today MGS works around this on the client side by checking for a change in the image after completing a poll loop: https://github.com/oxidecomputer/omicron/blob/a653803c51ee8d517da8b4b75ff09165323d017e/gateway/src/metrics.rs#L587-L593)

jamesmunns: Chatted with @hawkw, whose opinion is "[I] would prefer we basically start adding the restart ID to all new mgs messages".

We may not obtain the restart ID until after the system has finished a certain stage of booting (very very early), in which case I'll add a new error variant that covers this, probably not specific to these messages so we can re-use it for other things.

gateway-sp-comms/src/single_sp.rs
jamesmunns: Made it file-private

gateway-sp-comms/src/single_sp.rs
jamesmunns: Added

gateway-sp-comms/src/single_sp.rs
jamesmunns: Moved to file-private const

faux-mgs/src/main.rs
jamesmunns: Opened https://github.com/oxidecomputer/ipcc-rs/issues/12

jgallagher (approved):
LGTM! Thanks.

PR #496, opened 2026-06-01
Add `NullStr` type

jamesmunns:
This is a private type that is intended to be shared across wire types that act as a null-padded fixed length container format.

This is intended to be used in a follow up PR for supporting https://github.com/oxidecomputer/hubris/issues/2463, but this is separated to make reviewing this helper easier. Split out of https://github.com/oxidecomputer/management-gateway-service/tree/james/pmbus-please.

### DIFF
--- a/gateway-messages/src/lib.rs
+++ b/gateway-messages/src/lib.rs
@@ -279,8 +279,273 @@ impl From<UpdateId> for uuid::Uuid {
     }
 }
 
+mod nullstr {
+    use super::*;
+
+    /// A reusable type that implements a fixed-max-capacity string where the
+    /// unused capacity is always guaranteed to be zero-filled. This is used
+    /// for strings sent over the wire using [`hubpack`], which requires fixed
+    /// size items.
+    ///
+    /// This type does NOT guarantee that the contents are null-*terminated*,
+    /// only null-*padded*. If the contained string is exactly the length of the
+    /// container, there will be no null bytes present.
+    ///
+    /// It is not intended for this type to be part of the public API, see
+    /// [`SpComponent`] for usage guidance.
+    #[derive(Clone, Copy, PartialEq, Eq, Hash, SerializedSize)]
+    pub struct NullStr<const N: usize> {
+        contents: [u8; N],
+    }
+
+    impl<const N: usize> NullStr<N> {
+        /// Create a [`NullStr`] from the given slice.
+        ///
+        /// This function has some Interesting Details:
+        ///
+        /// 1. If `src` exceeds the length of `N`, only the first `N` bytes will
+        ///    be copied in
+        /// 2. If `src` contains null bytes, these will be copied in
+        ///
+        /// You might ask yourself, why isn't this unsafe? Why doesn't it
+        /// violate any invariants? Well, we implement `Deserialize`, which
+        /// means we might obtain garbage off the wire anyway! So, morally, this
+        /// is not particularly worse to support.
+        ///
+        /// This is *primarily* intended to be done when re-magicking from a
+        /// code-generated string.
+        pub fn from_bstr_unchecked(src: &[u8]) -> Self {
+            let mut buf = [0u8; N];
+            buf.iter_mut().zip(src.iter()).for_each(|(w, r)| *w = *r);
+            Self { contents: buf }
+        }
+
+        /// Interpret the content as a UTF-8 string.
+        ///
+        /// Our current expectation of contents is that this should never
+        /// fail (i.e., we're always storing contents as UTF-8 strings), but
+        /// because we reconstitute components from network messages we still
+        /// need to check.
+        pub fn as_str(&self) -> Option<&str> {
+            str::from_utf8(self.as_bstr()).ok()
+        }
+
+        /// Get the raw "binary string" version of the nullstr, e.g. the
+        /// contents prior to the first null byte, or the entire string if no
+        /// null bytes are present.
+        ///
+        /// This is guaranteed to never contain a `0`.
+        pub fn as_bstr(&self) -> &[u8] {
+            // `next()` can only be "None" if `N == 0`. Otherwise, split will
+            // always yield at least item.
+            self.contents.split(|x| *x == 0).next().unwrap_or(&[])
+        }
+
+        /// Const function to create a zero-padded [`NullStr`] from a given
+        /// [`str`].
+        ///
+        /// Panics if `src` exceeds the capacity of `N`. This should generally
+        /// only be used in const context where panics become compilation
+        /// errors.
+        pub const fn from_const(src: &str) -> Self {
+            let mut buf = [0u8; N];
+            assert!(src.len() <= N);
+            let mut idx = 0;
+            while idx < src.len() {
+                buf[idx] = src.as_bytes()[idx];
+                idx += 1;
+            }
+
+            Self { contents: buf }
+        }
+
+        /// Interpret the contents as a UTF-8 string in a `const` context,
+        /// panicking if the string is not valid.
+        ///
+        /// This function should only be used in const contexts when the caller
+        /// knows the contents is valid (e.g., one of this type's associated
+        /// constants); for contents names parsed or constructed at runtime,
+        /// prefer [`SpComponent::as_str()`] which performs runtime validation.
+        pub const fn const_as_str(&self) -> &str {
+            // const-equivalent of
+            // ```
+            // let n =
+            //    self.contents.iter().position(|&c| c == 0)
+            //      .unwrap_or(Self::MAX_ID_LENGTH);
+            // ```
+            let mut n = 0;
+            while n < self.contents.len() {
+                if self.contents[n] == 0 {
+                    break;
+                }
+                n += 1;
+            }
+
+            // const-equivalent of `let s = &self.contents[..n]`.
+            //
+            // SAFETY: We really want to say `&self.contents[..n]` here, but
+            // we're not allowed to use the indexing operator inside a
+            // `const fn`. We know from the loop above that
+            // `n <= self.contents.len()`, turning the following into a manual
+            // `&self.contents[..n]` without a bounds check.
+            let s = unsafe {
+                core::slice::from_raw_parts(self.contents.as_ptr(), n)
+            };
+
+            // const-equivalent of `str::from_utf8(s).unwrap_lite()`.
+            match str::from_utf8(s) {
+                Ok(s) => s,
+                Err(_) => panic!("invalid NullStr (not a utf8 string)"),
+            }
+        }
+
+        /// Obtain a view of the internal null-padded array of bytes.
+        pub fn contents(&self) -> &[u8; N] {
+            &self.contents
+        }
+    }
+
+    impl<const N: usize> Serialize for NullStr<N> {
+        fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+        where
+            S: serde::Serializer,
+        {
+            // NOTE: We are implicitly performing an action here similar to
+            // `serde(transparent)`, serializing the inner field directly rather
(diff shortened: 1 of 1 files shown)


gateway-messages/src/lib.rs
```diff
@@ -279,91 +279,280 @@ impl From<UpdateId> for uuid::Uuid {
...
+
+        /// Get the raw "binary string" version of the nullstr, e.g. the contents
+        /// prior to the first null byte, or the entire string if no null bytes
+        /// are present.
+        ///
+        /// This is guaranteed to never contain a `0`.
+        pub fn as_bstr(&self) -> &[u8] {
+            // `next()` can only be "None" if `N == 0`. Otherwise, split will
+            // always yield at least item.
+            self.contents.split(|x| *x == 0).next().unwrap_or(&[])
```
hawkw, line 341: this implementation is cute!

i do note that if we wanted to potentially save a few cycles, we might instead want an implementation that iterates over the characters starting at the *end* of the string until we find the first _non-null_ character, and split there, rather than iterating from the beginning until we find the first null, as `split` will. in practice, all the `NullStr`s we are working with are short enough that this probably doesn't matter a ton, but it felt like it would be worth pointing out.

gateway-messages/src/lib.rs
```diff
@@ -279,91 +279,280 @@ impl From<UpdateId> for uuid::Uuid {
...
+        pub fn as_str(&self) -> Option<&str> {
+            str::from_utf8(self.as_bstr()).ok()
+        }
+
+        /// Get the raw "binary string" version of the nullstr, e.g. the contents
+        /// prior to the first null byte, or the entire string if no null bytes
+        /// are present.
+        ///
+        /// This is guaranteed to never contain a `0`.
+        pub fn as_bstr(&self) -> &[u8] {
```
hawkw, line 338: personally, i might have just called this `as_bytes()`, but 🤷‍♀️

gateway-messages/src/lib.rs
```diff
@@ -279,91 +279,280 @@ impl From<UpdateId> for uuid::Uuid {
...
-        self.id.serialize(serializer)
-    }
-}
+        /// Interpret the content as a human-readable string.
+        ///
+        /// Our current expectation of contents is that this should never
+        /// fail (i.e., we're always storing contents as human-readable
+        /// strings), but because we reconstitute components from network messages
+        /// we still need to check.
+        pub fn as_str(&self) -> Option<&str> {
```
hawkw, line 329: i feel like this could return the `str::FromUtf8Error` rather than an `Option`?

gateway-messages/src/lib.rs
```diff
@@ -279,91 +279,280 @@ impl From<UpdateId> for uuid::Uuid {
...
-        // hubpack expects from serde's derived impl.
-        self.id.serialize(serializer)
-    }
-}
+        /// Interpret the content as a human-readable string.
+        ///
+        /// Our current expectation of contents is that this should never
+        /// fail (i.e., we're always storing contents as human-readable
+        /// strings), but because we reconstitute components from network messages
+        /// we still need to check.
```
hawkw: nit: I think "human-readable" is both kind of vague and not strictly accurate here. I think of "human-readable" as a much higher-level characteristic that separates e.g. a pretty-printed `Display` implementation from minified JSON, both of which are UTF-8 strings...and a string encoding that isn't UTF-8, such as UTF-16, ASCII, or even EBCDIC, could also contain human-readable data.

```suggestion
        /// Interpret the content as a UTF-8 string.
        ///
        /// Our current expectation of contents is that this should never
        /// fail (i.e., we're always storing contents as UTF-8 strings),
        /// but because we reconstitute components from network messages,
        /// we still need to check.
```

gateway-messages/src/lib.rs
```diff
@@ -279,91 +279,280 @@ impl From<UpdateId> for uuid::Uuid {
...
+            Self { contents: buf }
+        }
+
+        /// Interpret the contents as a human-readable string in a `const`
+        /// context, panicking if the string is not human readable.
+        ///
+        /// This function should only be used in const contexts when the caller
+        /// knows the contents is valid (e.g., one of this type's associated
+        /// constants); for contents names parsed or constructed at runtime, prefer
+        /// [`SpComponent::as_str()`] which performs runtime validation.
```
hawkw:
```suggestion
        /// Interpret the contents as a UTF-8 string in a `const`
        /// context, panicking if the string is not UTF-8.
        ///
        /// This function should only be used in const contexts when the caller
        /// knows the contents is valid (e.g., one of this type's associated
        /// constants); for contents names parsed or constructed at runtime, prefer
        /// [`SpComponent::as_str()`] which performs runtime validation.
```

gateway-messages/src/lib.rs
```diff
@@ -279,91 +279,280 @@ impl From<UpdateId> for uuid::Uuid {
...
+        fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+        where
+            S: serde::Serializer,
+        {
+            // NOTE: We are implicitly performing an action here similar to `serde(transparent)`,
+            // serializing the inner field directly rather than as a struct field. This is aesthetically
+            // pleasing for JSON, and not impactful for `hubpack`.
+            //
+            // If we're serializing to a human-readable form (e.g., `faux-mgs --json
+            // output`), serialize ourself as a string....
```
hawkw: nit: can we wrap this as 80 columns?

gateway-messages/src/lib.rs
```diff
@@ -279,91 +279,280 @@ impl From<UpdateId> for uuid::Uuid {
...
+        where
+            D: serde::Deserializer<'de>,
+        {
+            use serde::de::Visitor;
+            // NOTE: We are implicitly performing an action here similar to `serde(transparent)`,
+            // deserializing the inner field directly rather than as a struct field. This is aesthetically
+            // pleasing for JSON, and not impactful for `hubpack`.
+            //
+            // Inverse of our serialize method: if we're deserializing from a
+            // human-readable form, deserialize a string...
```
hawkw: can we wrap this too?

gateway-messages/src/lib.rs
```diff
@@ -441,34 +639,36 @@ impl SpComponent {
...
+    /// This is *primarily* intended to be done when re-magicking from a code-generated string.
+    #[inline]
+    pub fn from_bstr_unchecked(src: &[u8]) -> Self {
+        Self { id: nullstr::NullStr::from_bstr_unchecked(src) }
+    }
+
+    #[inline]
+    pub const fn from_const(val: &str) -> Self {
+        Self { id: nullstr::NullStr::from_const(val) }
+    }
```
hawkw, line 678: huh, this all ends up a bit boilerplate-y; I thought about saying "can't we have some kinda `From<T> where T: Into<NullStr>`, but I guess we can't, because the whole point of the different constructors is to have some be const-callable so it doesn't really work out. This still seem sworth it, just a bit of a bother.

gateway-messages/src/lib.rs
```diff
@@ -279,91 +279,280 @@ impl From<UpdateId> for uuid::Uuid {
...
+            // ... otherwise, serialize our id array directly, which matches what
+            // hubpack expects from serde's derived impl.
+            //
+            // NOTE: We use `serde_big_array` as we are const-generic over the
+            // length, and serde's implementation is *not* generic, and only provides
+            // manual impls up to N=32.
+            <[u8; N] as serde_big_array::BigArray<'static, u8>>::serialize(
+                &self.contents,
+                serializer,
+            )
```
hawkw: Huh! I realize that this is what we were already doing, but it feels low-key goofy. It feels like using `serialize_str` is actually more efficient because it wouldn't have encoded the padding bytes. WITH THAT SAID, I think that might be a breaking wire format change, so we have to do it this way. Oh well.

gateway-messages/src/lib.rs
```diff
@@ -279,91 +279,280 @@ impl From<UpdateId> for uuid::Uuid {
...
+        /// This function has some Interesting Details:
+        ///
+        /// 1. If `src` exceeds the length of `N`, only the first `N` bytes will be copied in
+        /// 2. If `src` contains null bytes, these will be copied in
+        ///
+        /// You might ask yourself, why isn't this unsafe? Why doesn't it violated any invariants?
+        /// Well, we implement `Deserialize`, which means we might obtain garbage off the wire
+        /// anyway! So, morally, this is not particularly worse to support.
+        ///
+        /// This is *primarily* intended to be done when re-magicking from a code-generated string.
```
hawkw: I think some of these ought to be wrapped at the 80th column

hawkw (commented):
Thanks for doing this! As I mentioned, it would be nice if, eventually, this could be merged with `FixedStr`, but it seems good for now.

I left some relatively minor notes; overall, the implementation looks solid.

gateway-messages/src/lib.rs
jamesmunns: Thanks! I hadn't enabled the ruler for Zed yet, but this should be fixed now!

gateway-messages/src/lib.rs
jamesmunns: Interesting, do you think that this would just stride better while looking for "not zero"?

On embedded I don't think it would be faster, but with the joys of SIMD I could totally believe it would be.

gateway-messages/src/lib.rs
jamesmunns: yeah, I specifically chose `bstr` because it doesn't necessarily give you the *whole* buffer, but semantically the same as `as_str().as_bytes()`. If I could make that more clear, I can!

gateway-messages/src/lib.rs
jamesmunns: It could! But this code was already there. I could change it if you think that would be a good idea though!

gateway-messages/src/lib.rs
jamesmunns: Will fix!

gateway-messages/src/lib.rs
jamesmunns: I think that would actually be wrong WRT `hubpack`: if we tried to serialize as a string it would break because hubpack doesn't support it.

gateway-messages/src/lib.rs
hawkw: yeah, that's fair, I think `as_bstr()` is a bit clearer about that. carry on.

gateway-messages/src/lib.rs
hawkw: I was really just thinking about the number of iterations to find the split point. This is based purely on the guess that most `NullStr`s will contain more non-null characters than null characters...which, after I think about it slightly longer, is actually _not_ the case for `SpComponentId`s, where most of them are refdes that are going to be 3-4 character strings like `U69` or `U420`. So actually, the typical `SpComponentId` will do substantially fewer iterations starting from the beginning rather than the end. 

I think for power rails, which are longer strings, this wouldn't be the case if the buffer was also 16 characters there, but it's 32, so I actually think the average `NullStr` is more nulls than non-nulls in either case, and I'm just straight up wrong and the current thing is better. NEVER MIND LOL.

gateway-messages/src/lib.rs
hawkw: Oh, it will just error out if you give it a string, won't it? Never mind!

gateway-messages/src/lib.rs
hawkw: :woman_shrugging:

hawkw (approved):
okay, this seems good to me! it might be worth testing that a `faux-mgs` built from this branch can happily call into the `inventory` and `component-details` APIs on a SP running the current release Hubris image, to ensure we've not accidentally broken wire compatibility someplace? i feel like that's unlikely based on the change but it is probably still worth checking.

PR #497, opened 2026-06-01
Add PMBus status retrieval to MGS API

jamesmunns:
Follow-on PR to #496.

Part of https://github.com/oxidecomputer/hubris/issues/2463.

### DIFF
--- a/faux-mgs/src/main.rs
+++ b/faux-mgs/src/main.rs
@@ -536,7 +536,7 @@ enum Command {
     GetHostFlashHash {
         slot: u16,
     },
-    /// Get the PMBus Status values for a given Power Rail
+    /// Get the PMBus status registers for a given Power Rail
     GetPmbusStatus {
         /// Power rail name, as referred to in the App TOML
         #[clap(value_parser = parse_power_rail_name)]
@@ -2480,49 +2480,49 @@ async fn run_command(
         Command::GetPmbusStatus { rail } => {
             let result = sp.get_pmbus_status(rail).await?;
             if json {
-                Ok(Output::Json(json!(result)))
-            } else {
-                let name = result.rail.as_str().unwrap();
-                let PmbusStatus {
-                    status_word,
-                    status_vout,
-                    status_iout,
-                    status_temperature,
-                    status_cml,
-                    status_other,
-                    status_input,
-                    status_mfr_specific,
-                    status_fans_1_2,
-                    status_fans_3_4,
-                } = result.status;
-
-                let mut lines = vec![];
-                lines.push(format!("Rail '{name}':"));
-                lines.push(format!("WORD          0x{status_word:04x}"));
-                let data = [
-                    ("VOUT          ", status_vout),
-                    ("IOUT          ", status_iout),
-                    ("TEMPERATURE   ", status_temperature),
-                    ("CML           ", status_cml),
-                    ("OTHER         ", status_other),
-                    ("INPUT         ", status_input),
-                    ("MFR_SPECIFIC  ", status_mfr_specific),
-                    ("FANS_1_2      ", status_fans_1_2),
-                    ("FANS_3_4      ", status_fans_3_4),
-                ];
-
-                use std::fmt::Write;
-                for (n, v) in data {
-                    let mut l = n.to_string();
-                    match v {
-                        Ok(b) => write!(&mut l, "0x{b:02x}")?,
-                        Err(e) => write!(&mut l, "Err({e:?})")?,
-                    }
-                    lines.push(l);
-                }
-
-                Ok(Output::Lines(lines))
+                return Ok(Output::Json(json!(result)));
             }
+
+            let name = result.rail.as_str().unwrap();
+            let PmbusStatus {
+                status_word,
+                status_vout,
+                status_iout,
+                status_temperature,
+                status_cml,
+                status_other,
+                status_input,
+                status_mfr_specific,
+                status_fans_1_2,
+                status_fans_3_4,
+            } = result.status;
+
+            let mut lines = vec![];
+            lines.push(format!("Rail '{name}':"));
+            lines.push(format!("WORD          0x{status_word:04x}"));
+            let data = [
+                ("VOUT          ", status_vout),
+                ("IOUT          ", status_iout),
+                ("TEMPERATURE   ", status_temperature),
+                ("CML           ", status_cml),
+                ("OTHER         ", status_other),
+                ("INPUT         ", status_input),
+                ("MFR_SPECIFIC  ", status_mfr_specific),
+                ("FANS_1_2      ", status_fans_1_2),
+                ("FANS_3_4      ", status_fans_3_4),
+            ];
+
+            use std::fmt::Write;
+            for (n, v) in data {
+                let mut l = n.to_string();
+                match v {
+                    Ok(b) => write!(&mut l, "0x{b:02x}")?,
+                    Err(e) => write!(&mut l, "Err({e:?})")?,
+                }
+                lines.push(l);
+            }
+
+            Ok(Output::Lines(lines))
         }
     }
 }
--- a/faux-mgs/src/main.rs
+++ b/faux-mgs/src/main.rs
@@ -25,6 +25,8 @@ use gateway_messages::IgnitionCommand;
 use gateway_messages::LedComponentAction;
 use gateway_messages::MonorailComponentAction;
 use gateway_messages::MonorailComponentActionResponse;
+use gateway_messages::PmbusStatus;
+use gateway_messages::PowerRailName;
 use gateway_messages::PowerState;
 use gateway_messages::ROT_PAGE_SIZE;
 use gateway_messages::RotBootInfo;
@@ -533,6 +535,12 @@ enum Command {
     },
     GetHostFlashHash {
         slot: u16,
+    },
+    /// Get the PMBus Status values for a given Power Rail
+    GetPmbusStatus {
+        /// Power rail name, as referred to in the App TOML
+        #[clap(value_parser = parse_power_rail_name)]
+        rail: PowerRailName,
     },
 }
 
@@ -1087,6 +1095,11 @@ fn parse_tlvc_key(key: &str) -> Result<[u8; 4]> {
 fn parse_sp_component(component: &str) -> Result<SpComponent> {
     SpComponent::try_from(component)
         .map_err(|_| anyhow!("invalid component name: {component}"))
+}
+
+fn parse_power_rail_name(name: &str) -> Result<PowerRailName> {
+    PowerRailName::try_from(name)
+        .map_err(|_| anyhow!("invalid power rail name: {name}"))
 }
 
 fn parse_bulk_targets_shim(s: &str) -> Result<IgnitionBulkTargets> {
@@ -2463,6 +2476,53 @@ async fn run_command(
         Command::GetHostFlashHash { slot } => {
             let result = sp.get_host_flash_hash(slot).await?;
             Ok(Output::Lines(vec![format!("{result:x?}")]))
+        }
+        Command::GetPmbusStatus { rail } => {
+            let result = sp.get_pmbus_status(rail).await?;
+            if json {
+                Ok(Output::Json(json!(result)))
+            } else {
+                let name = result.rail.as_str().unwrap();
+                let PmbusStatus {
+                    status_word,
+                    status_vout,
+                    status_iout,
+                    status_temperature,
+                    status_cml,
+                    status_other,
+                    status_input,
+                    status_mfr_specific,
+                    status_fans_1_2,
+                    status_fans_3_4,
+                } = result.status;
+
+                let mut lines = vec![];
+                lines.push(format!("Rail '{name}':"));
+                lines.push(format!("WORD          0x{status_word:04x}"));
+                let data = [
+                    ("VOUT          ", status_vout),
+                    ("IOUT          ", status_iout),
+                    ("TEMPERATURE   ", status_temperature),
+                    ("CML           ", status_cml),
+                    ("OTHER         ", status_other),
+                    ("INPUT         ", status_input),
+                    ("MFR_SPECIFIC  ", status_mfr_specific),
+                    ("FANS_1_2      ", status_fans_1_2),
+                    ("FANS_3_4      ", status_fans_3_4),
+                ];
+
+                use std::fmt::Write;
+                for (n, v) in data {
+                    let mut l = n.to_string();
+                    match v {
+                        Ok(b) => write!(&mut l, "0x{b:02x}")?,
+                        Err(e) => write!(&mut l, "Err({e:?})")?,
+                    }
+                    lines.push(l);
+                }
+
+                Ok(Output::Lines(lines))
+            }
         }
     }
 }
--- a/gateway-messages/src/lib.rs
+++ b/gateway-messages/src/lib.rs
@@ -649,6 +649,71 @@ impl TryFrom<&str> for SpComponent {
     }
 }
 
+/// Identifier for a single power rail.
+#[derive(
+    Clone, Copy, PartialEq, Eq, Hash, SerializedSize, Serialize, Deserialize,
+)]
+#[serde(transparent)]
+pub struct PowerRailName {
+    name: nullstr::NullStr<{ Self::MAX_NAME_LENGTH }>,
+}
+
+impl core::fmt::Display for PowerRailName {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        self.name.fmt(f)
+    }
+}
+
+impl PowerRailName {
+    /// Maximum number of bytes for a Power Rail Name.
+    pub const MAX_NAME_LENGTH: usize = 32;
+
+    /// Interpret the power rail name as a human-readable string.
+    ///
+    /// Our current expectation of power rail names is that this should never
+    /// fail (i.e., we're always storing power rail names as human-readable
+    /// strings), but because we reconstitute components from network messages
+    /// we still need to check.
+    #[inline]
+    pub fn as_str(&self) -> Option<&str> {
+        self.name.as_str()
+    }
+
+    /// Interpret the power rail name as a human-readable string in a `const`
+    /// context, panicking if the string is not human readable.
+    ///
+    /// This function should only be used in const contexts when the caller
+    /// knows the component is valid (e.g., one of this type's associated
+    /// constants); for component names parsed or constructed at runtime, prefer
+    /// [`PowerRailName::as_str()`] which performs runtime validation.
+    #[inline]
+    pub const fn const_as_str(&self) -> &str {
+        self.name.const_as_str()
+    }
+
+    #[inline]
+    pub const fn from_const(val: &str) -> Self {
+        Self { name: nullstr::NullStr::from_const(val) }
+    }
+
+    #[inline]
+    pub fn id(&self) -> &[u8; Self::MAX_NAME_LENGTH] {
+        self.name.contents()
+    }
+}
+
+impl fmt::Debug for PowerRailName {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let mut debug = f.debug_struct("PowerRailName");
+        if let Some(s) = self.as_str() {
+            debug.field("name", &s);
+        } else {
+            debug.field("name", self.name.contents());
+        }
+        debug.finish()
+    }
+}
+
 /// Minimum guaranteed space for trailing data in a single packet.
 ///
 /// Depending on the [`Message`] payload, there may be more space for trailing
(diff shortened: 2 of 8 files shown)


gateway-messages/src/sp_to_mgs.rs
```diff
@@ -186,6 +187,17 @@ pub enum SpResponse {
...
     ComponentPersistentSlot(u16),
+
+    /// PMBus status of a given rail
+    PmbusStatus(PMBusStatusResponse),
+}
+
+#[derive(
+    Clone, Copy, Debug, PartialEq, Serialize, Deserialize, SerializedSize,
+)]
+pub struct PMBusStatusResponse {
```
hawkw: Nit: I think this really ought to be
```suggestion
pub struct PmbusStatusResponse {
```
since the generally-agreed-upon Rust style is usually for acronyms and similar to be treated as one "word" when camel-cased (and the result types are `PmbusStatus` and `PmbusStatusError`, and the response enum variant is `PmbusStatus`...)

gateway-messages/src/mgs_to_sp.rs
```diff
@@ -241,6 +242,9 @@ pub enum MgsRequest {
     /// current default slot as persisted in non-volatile memory. This may be
     /// different than the current active slot (see `ComponentGetActiveSlot`).
     ComponentGetPersistentSlot(SpComponent),
+
+    /// Request the STATUS registers of a PMBus device, indexed by Power Rail Name
+    GetPMBusStatus(PowerRailName),
```
hawkw: I think this should probably be 
```suggestion
    GetPmbusStatus(PowerRailName),
```
for consistency with all the other names?

wireshark/protofields.lua
```diff
@@ -95,6 +95,7 @@ M.mgs_request.names = {
     [48] = "StartHostFlashHash",
     [49] = "GetHostFlashHash",
     [50] = "ComponentGetPersistentSlot",
+    [51] = "GetPmBusStatus",
```
hawkw:
```suggestion
    [51] = "GetPmbusStatus",
```

wireshark/protofields.lua
```diff
@@ -148,6 +149,7 @@ M.mgs_request.handlers = {
     [48] = "dissect_start_host_flash_hash",
     [49] = "dissect_get_host_flash_hash",
     [50] = "dissect_component_get_persistent_slot",
+    [51] = "dissect_get_pm_bus_status",
```
hawkw: it's weird that we are spelling it `pm_bus_status` here but in the response handler we are spelling it `pmbus_status` --- let's make this consistent? (also i think it should always be spelled `pmbus`)

gateway-messages/src/sp_to_mgs.rs
```diff
@@ -712,6 +724,27 @@ pub struct TlvPage {
     pub total: u32,
 }
 
+#[derive(
+    Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, SerializedSize,
+)]
+pub enum PmbusStatusError {}
```
hawkw: i presume this will basically end up being an `i2c::ResponseCode`?

gateway-messages/src/lib.rs
```diff
@@ -704,6 +669,71 @@ impl TryFrom<&str> for SpComponent {
...
+
+impl core::fmt::Display for PowerRailName {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        self.name.fmt(f)
+    }
+}
+
+impl PowerRailName {
+    /// Maximum number of bytes for a Power Rail Name.
+    pub const MAX_NAME_LENGTH: usize = 32;
```
hawkw, line 743: maybe the API docs should explicitly note that `from_const` and friends will panic if the string is longer than this?

gateway-messages/src/lib.rs
```diff
@@ -704,6 +669,71 @@ impl TryFrom<&str> for SpComponent {
...
+    /// fail (i.e., we're always storing power rail names as human-readable
+    /// strings), but because we reconstitute components from network messages
+    /// we still need to check.
+    #[inline]
+    pub fn as_str(&self) -> Option<&str> {
+        self.name.as_str()
+    }
+
+    /// Interpret the power rail name as a human-readable string in a `const`
+    /// context, panicking if the string is not human readable.
```
hawkw:
```suggestion
    /// Interpret the power rail name as a UTF-8 string in a `const`
    /// context, panicking if the string is not UTF-8.
```

gateway-messages/src/lib.rs
```diff
@@ -704,6 +669,71 @@ impl TryFrom<&str> for SpComponent {
...
+impl PowerRailName {
+    /// Maximum number of bytes for a Power Rail Name.
+    pub const MAX_NAME_LENGTH: usize = 32;
+
+    /// Interpret the power rail name as a human-readable string.
+    ///
+    /// Our current expectation of power rail names is that this should never
+    /// fail (i.e., we're always storing power rail names as human-readable
+    /// strings), but because we reconstitute components from network messages
+    /// we still need to check.
```
hawkw:
```suggestion
    /// Interpret the power rail name as a UTF-8 string.
    ///
    /// Our current expectation of power rail names is that this should never
    /// fail (i.e., we're always storing power rail names as UTF-8 strings),
    /// but because we reconstitute components from network messages, we still
    /// need to check.
```

gateway-messages/src/lib.rs
```diff
@@ -704,6 +669,71 @@ impl TryFrom<&str> for SpComponent {
...
+        self.name.const_as_str()
+    }
+
+    #[inline]
+    pub const fn from_const(val: &str) -> Self {
+        Self { name: nullstr::NullStr::from_const(val) }
+    }
+
+    #[inline]
+    pub fn id(&self) -> &[u8; Self::MAX_NAME_LENGTH] {
```
hawkw: since this type is `PowerRailName` and not `SpComponentId`, I might name this something different...

hawkw (commented):
Overall, this seems like a good start, I left a few suggestions.

Also, it would be nice to add support for this to `faux-mgs`. If you want to be fancy, it would be really cool if the non-JSON output for the `faux-mgs` command pretty-printed the decoded value of each register using the `pmbus` crate, in addition to printing the raw value of the register. That would probably make life much easier for the folks who end up using the MGS command to debug stuff. Feel free to steal from [the pretty-printer I wrote for `erebor`](https://github.com/oxidecomputer/erebor/blob/48512bf970474ff0fd0868c833fd504c8d169064/src/lib.rs#L232-L346) and/or the `humility pmbus` command, as you see fit (though note that `humility pmbus`' pretty-printer is way more complex, as it also decodes PMBus registers that are _not_ always just bitflags, as the status registers are)...

gateway-messages/src/sp_to_mgs.rs
jamesmunns: I was curious and went to check in the `hubris` repo with a quick project grep of .rs files, and today it's *about* a 50:50 split between `Pmbus` and `PMBus` (unsurprisingly no `PmBus`!). I'm pretty sure I got this casing from `build_i2c`, which uses `PMBus` mostly it seems.

To be fair, there's quite a few false positives for `PMBus`: it shows up in comments as well, though quite a bit in code still.

We might want to make a pass and homogenize it some day, I'll switch this PR over to `Pmbus` though.

wireshark/protofields.lua
jamesmunns: I didn't hand-edit this, it got filled in by the "bless" step of some kind of snapshot-like testing.

gateway-messages/src/sp_to_mgs.rs
jamesmunns: Pretty close! Plus maybe one or two extra variants. I still need to fill these in, they are unwraps in the current hubris code.

wireshark/protofields.lua
hawkw: oh cool, i didn't realize we were doing that for these!

faux-mgs/src/main.rs
```diff
@@ -534,6 +536,12 @@ enum Command {
     GetHostFlashHash {
         slot: u16,
     },
+    /// Get the PMBus Status values for a given Power Rail
```
hawkw: unimportant turbo-nits:
```suggestion
    /// Get the PMBus status registers for a given power rail.
```

faux-mgs/src/main.rs
```diff
@@ -2464,6 +2477,53 @@ async fn run_command(
...
+
+                use std::fmt::Write;
+                for (n, v) in data {
+                    let mut l = n.to_string();
+                    match v {
+                        Ok(b) => write!(&mut l, "0x{b:02x}")?,
+                        Err(e) => write!(&mut l, "Err({e:?})")?,
+                    }
+                    lines.push(l);
+                }
```
hawkw: this is cute! in future it might be nice to hook this onto the `pmbus` crate to decode the bit patterns, but i very much do not care about doing it in _this_ PR.

faux-mgs/src/main.rs
```diff
@@ -2464,6 +2477,53 @@ async fn run_command(
             let result = sp.get_host_flash_hash(slot).await?;
             Ok(Output::Lines(vec![format!("{result:x?}")]))
         }
+        Command::GetPmbusStatus { rail } => {
+            let result = sp.get_pmbus_status(rail).await?;
+            if json {
+                Ok(Output::Json(json!(result)))
+            } else {
```
hawkw: nit, up to you, but, we could save a level of indentation (a thing which I am weirdly sensitive about due to some kind of reason) if we did 
```suggestion
            if json {
                return Ok(Output::Json(json!(result)));
            }
```
here?

gateway-messages/src/sp_to_mgs.rs
```diff
@@ -712,6 +724,72 @@ pub struct TlvPage {
...
+}
+
+#[derive(
+    Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, SerializedSize,
+)]
+pub enum PmbusStatusReadError {
+    /// Failed while attempting to communicate with I2C driver
+    DriverReadFailed { response_code: u8 },
+    /// Received unexpected data on the I2C bus
+    InvalidBusData,
```
hawkw: if we stop trying to interpret the status bits and then turn the m back into `u8`s, i think we can just get rid of this?

gateway-messages/src/sp_to_mgs.rs
```diff
@@ -1283,6 +1362,7 @@ impl fmt::Display for SpError {
             Self::Monorail(e) => write!(f, "monorail: {}", e),
             Self::Dump(e) => write!(f, "dump: {}", e),
             Self::Hf(e) => write!(f, "hf: {}", e),
+            Self::PmbusStatus(e) => write!(f, "pmbus stat: {}", e),
```
hawkw:
```suggestion
            Self::PmbusStatus(e) => write!(f, "pmbus status: {}", e),
```

gateway-messages/src/lib.rs
```diff
@@ -715,6 +715,121 @@ impl TryFrom<&str> for SpComponent {
...
+    }
+}
+
+impl fmt::Debug for PowerRailName {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let mut debug = f.debug_struct("PowerRailName");
+        if let Some(s) = self.as_str() {
+            debug.field("name", &s);
+        } else {
+            debug.field("name", self.name.contents());
```
hawkw: i kinda wonder if `NullStr` should just have a `Debug` impl that does this, so that we can derive `Debug` for `PowerRailName` and `SpComponent`?

hawkw (approved):
This looks good. I had some smallish suggestions.

gateway-messages/src/lib.rs
jamesmunns: Will do this, that's a good idea.

faux-mgs/src/main.rs
```diff
@@ -534,6 +536,12 @@ enum Command {
     GetHostFlashHash {
         slot: u16,
     },
+    /// Get the PMBus status registers for a given Power Rail
```
hawkw, line 539: turbo nit: i would probably not capitalize this.

gateway-messages/src/mgs_to_sp.rs
```diff
@@ -241,6 +242,9 @@ pub enum MgsRequest {
     /// current default slot as persisted in non-volatile memory. This may be
     /// different than the current active slot (see `ComponentGetActiveSlot`).
     ComponentGetPersistentSlot(SpComponent),
+
+    /// Request the STATUS registers of a PMBus device, indexed by Power Rail Name
```
hawkw: turbo-nit: i would either not capitalize this or i would link to the actual type, so either:
```suggestion
    /// Request the STATUS registers of a PMBus device, indexed by power rail name.
```
or
```suggestion
    /// Request the STATUS registers of a PMBus device, indexed by [`PowerRailName`].
```

gateway-messages/src/sp_to_mgs.rs
```diff
@@ -712,6 +724,91 @@ pub struct TlvPage {
...
+    FailedStatusWord(PmbusStatusReadError),
+}
+
+impl fmt::Display for PmbusStatusError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let msg = match self {
+            PmbusStatusError::UnknownRail => "Unknown PMBus Rail Name",
+            PmbusStatusError::FailedStatusWord(reason) => {
+                return write!(f, "Failed to read STATUS_WORD: {reason}");
+            }
```
hawkw: turbo nit: I think that in general, we consider it better style not to capitalize the first characters error messages, since they may get formatted with another string in front of them as a source chain.
```suggestion
            PmbusStatusError::UnknownRail => "unknown PMBus rail name",
            PmbusStatusError::FailedStatusWord(reason) => {
                return write!(f, "failed to read STATUS_WORD: {reason}");
            }
```

hawkw: also, take it or leave it, but i think that the `UnknownRail` message could maybe be reworded a little bit to say, specifically, that the provided rail name is _not a PMBus device_, since there may be power rails named in a manifest which are not provided by PMBus devices.

gateway-messages/src/sp_to_mgs.rs
```diff
@@ -712,6 +724,91 @@ pub struct TlvPage {
...
+        /// The raw I2C driver code reported when this query failed. This value
+        /// is not stable across versions of the SP firmware, and should only
+        /// be logged or used for interactive or post-mortem debugging.
+        /// Requires knowledge of the exact firmware revision to meaningfully
+        /// decode.
+        raw_response_code: u8,
+    },
+    /// This device does not support this status register, and it was not
+    /// queried by the SP.
+    Unsupported,
```
hawkw, line 769: nit: I might prefer
```suggestion
    UnsupportedRegister
```
for clarity, or something, although I guess it's _fairly_ obvious because this error only shows up for a specific register in the struct...

gateway-messages/src/sp_to_mgs.rs
```diff
@@ -712,6 +724,91 @@ pub struct TlvPage {
...
+                retry_hint,
+                raw_response_code,
+            } => {
+                let diagnosis = if !*retry_hint { "fatal" } else { "retry" };
+                return write!(
+                    f,
+                    "I2C Driver Error Code: {raw_response_code}, {diagnosis}"
+                );
+            }
+            PmbusStatusReadError::Unsupported => "Unsupported status field",
```
hawkw: nit: maybe:
```suggestion
                let diagnosis = if !*retry_hint { "fatal" } else { "retryable" };
                return write!(
                    f,
                    "I2C driver error code: {raw_response_code} ({diagnosis})"
                );
            }
            PmbusStatusReadError::Unsupported => "unsupported status field",
```

gateway-messages/src/sp_to_mgs.rs
```diff
@@ -712,6 +724,91 @@ pub struct TlvPage {
...
+        };
+
+        f.write_str(msg)
```
hawkw: to be quite honest, the `let msg` thing with the early return and so on feels a bit overengineered to me given that there's only one case where we set a value for `msg`. so we could maybe just do this:
```suggestion
        match self {
            PmbusStatusReadError::DriverReadFailed {
                retry_hint,
                raw_response_code,
            } => {
                let diagnosis = if !*retry_hint { "fatal" } else { "retry" };
                write!(
                    f,
                    "I2C Driver Error Code: {raw_response_code}, {diagnosis}"
                )
            }
            PmbusStatusReadError::Unsupported => f.write_str("Unsupported status field"),
        }
```

gateway-messages/src/sp_to_mgs.rs
```diff
@@ -1283,6 +1381,7 @@ impl fmt::Display for SpError {
             Self::Monorail(e) => write!(f, "monorail: {}", e),
             Self::Dump(e) => write!(f, "dump: {}", e),
             Self::Hf(e) => write!(f, "hf: {}", e),
+            Self::PmbusStatus(e) => write!(f, "pmbus status: {}", e),
```
hawkw:
```suggestion
            Self::PmbusStatus(e) => write!(f, "PMBus status: {}", e),
```

gateway-messages/src/lib.rs
hawkw: are we still planning to do something like this? or should we save that for a follow-up?

hawkw (approved):
still feel good about this, feel free to merge whenever you like

gateway-messages/src/lib.rs
jamesmunns: Ah shoot, lemme do this quick.

gateway-messages/src/lib.rs
```diff
@@ -516,23 +518,19 @@ mod nullstr {
...
     }
 
     impl<const N: usize> core::fmt::Display for NullStr<N> {
         fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
-            if let Some(s) = self.as_str() {
-                write!(f, "{s}")
-            } else {
-                write!(f, "{self:?}")
-            }
+            <Self as fmt::Debug>::fmt(self, f)
```
hawkw: hm, i'm actually not sure if i like this one, as the `Display` format will now put quotes around the string, which `Display` usually doesn't do. i'd prefer the `Display` impl to use `str`'s `Display` impl if it's a string, and print the `Debug` impl for the byte array otherwise. that means that the `Display` and `Debug` impls will still both need to have their own version of the `if`, but i think that's actually better if it gets us formatting that's more consistent with the rest of Rust...

gateway-messages/src/lib.rs
```diff
@@ -516,23 +518,19 @@ mod nullstr {
 
     impl<const N: usize> fmt::Debug for NullStr<N> {
         fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-            let mut debug = f.debug_struct("NullStr");
             if let Some(s) = self.as_str() {
-                debug.field("contents", &s);
+                f.write_char('"')?;
+                f.write_str(s)?;
+                f.write_char('"')
```
hawkw: hm, to be honest, i feel like we should probably just do
`fmt::Debug::fmt(self, s)` here; if we're going to the effort of putting quotes around it in the Debug format, we may as well also escape non-printable characters like the `Debug` impl for `str` will do.

gateway-messages/src/lib.rs
```diff
@@ -516,23 +518,19 @@ mod nullstr {
...
             } else {
-                debug.field("contents", &self.contents);
+                write!(f, "{:?}", self.contents)
```
hawkw: as an aside, i sure wish this would get us the byte array in hex rather than as an array of decimal numbers...but not worth trying to fix that here, it's fine.

gateway-messages/src/sp_to_mgs.rs
```diff
@@ -771,21 +771,22 @@ pub enum PmbusStatusReadError {
...
                     "I2C driver error code: {raw_response_code} ({diagnosis})"
-                );
+                )
             }
-            PmbusStatusReadError::Unsupported => "unsupported status field",
-        };
-
-        f.write_str(msg)
+            PmbusStatusReadError::Unsupported => {
+                f.write_str("unsupported status field")
```
hawkw: nit: might say "register" rather than "field" here, but it's up to you.

hawkw (commented):
okay, sorry to be pedantic about the new `Display` and `Debug` impls, but...

gateway-messages/src/lib.rs
```diff
@@ -516,22 +516,20 @@ mod nullstr {
...
     }
 
     impl<const N: usize> core::fmt::Display for NullStr<N> {
         fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
             if let Some(s) = self.as_str() {
-                write!(f, "{s}")
+                f.write_str(s)
             } else {
-                write!(f, "{self:?}")
+                write!(f, "{:?}", self.contents)
```
hawkw: should this maybe also be 
```suggestion
                write!(f, "{:02x?}", self.contents)
```

hawkw (approved):
let's goooooo! 🙌

PR #498, opened 2026-07-10
Bump version to 26

jamesmunns:
This should have been done for https://github.com/oxidecomputer/management-gateway-service/pull/497

CC #495, which will bump to v27

### DIFF
--- a/gateway-messages/tests/versioning/mod.rs
+++ b/gateway-messages/tests/versioning/mod.rs
@@ -31,6 +31,7 @@ mod v22;
 mod v23;
 mod v24;
 mod v25;
+mod v26;
 
 pub fn assert_serialized<T: Serialize + SerializedSize + std::fmt::Debug>(
     expected: &[u8],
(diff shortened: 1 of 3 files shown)


gateway-messages/tests/versioning/mod.rs
```diff
@@ -31,7 +31,9 @@ mod v22;
 mod v23;
 mod v24;
 mod v25;
+mod v26;
 
+#[track_caller]
```
jamesmunns, line 36: If you'd like, I can back this change out, but I think it's useful. It means we get errors like this:

```
thread 'versioning::v26::get_pmbus_status_error' panicked at gateway-messages/tests/versioning/v26.rs:121:9:
assertion `left == right` failed: incorrect serialization of PmbusStatus(UnknownRail)
  left: [123, 0]
 right: [39, 0]
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

instead of:

```
thread 'versioning::v26::get_pmbus_status_error' panicked at gateway-messages/tests/versioning/mod.rs:48:5:
assertion `left == right` failed: incorrect serialization of PmbusStatus(UnknownRail)
  left: [123, 0]
 right: [39, 0]
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

jgallagher (approved):

